PR #4, opened 2023-10-25
Refactoring

3nids:
This refactoring is mainly about moving files around in the desired structure according to the TEKSI WW datamodel structure + init strategy meeting from october 16th. In short:
* all table definition are moved to the changelogs folder
* all view + triggers are move to the app folder

For swmm, this means that all the views are moved to app (but named swmm_vw_*), only 1 table remains in a dedicated schema.

I have the feeling that we have too many schemas floating around bringing too much complexity.
Currently we have 1 table in import, 2 in network, 1 in swmm...so 3 schema for 4 tables!
The number of schemas makes any operation on the DB (migration/update, dumps, etc) a bit more cumbersome for no added value to me.

I would be in favor of limiting the number of schema to:
* `od` for VSA tables
* `vl` for VSA domains
* `sys` for system
* `additional` for anything else (for import/export, network or extension like swmm or anything else)
* `app` for the application

I think this would be applicable to any solution (drinkingwater, etc)

Other implications / questions:
* audit table has been dropped, any objection?
* do we really need a complex oid_generation? could we live with a simple UUID? This means less maintenance, better performance and the generation on client side is easy / possible (when editing offline for instance).
* there is a trigger to set the parent last modification date (https://github.com/teksi/wastewater/blob/e5622cbe35f6c9a36a2fbf1d22f45bda271a3bb8/datamodel/changelogs/0001/02_sys.sql#L10C25-L10C52). Is this needed?
* dictionnaries: I have left them on the side for now, this will be questionned when we evaluate the translation again
* what PostgreSQL minimum version do we want? I would suggest 11 (https://www.postgresql.org/support/versioning/)
* aquifier and planning zone tables have been prefixed with an underscore (because they're not part of VSA). I am not sure this is coherent accross the whole datamodel. I would suggest to move them to an "additional" schema (as suggested above).

General:
* Current CI creates dumps of the datamodel which are found as artifacts of the workflow. For instance: https://github.com/teksi/wastewater/actions/runs/6642617262
* I have added some auto formatting to the repository (pre-commit)


datamodel/.docker/.pg_service.conf
```diff
@@ -0,0 +1,5 @@
+[pg_teksi_wastewater]
+dbname=teksi_wastewater
+user=postgres
+password=postgres
+host=signalo
```
domi4484, line 5: rename to teksi something?

datamodel/.docker/init_db.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+DB_BASE_NAME=signalo
```
domi4484, line 3: Here as well

datamodel/app/view/__pycache__/vw_tww_reach.cpython-311.pyc
domi4484, line 1: Should not be in git

datamodel/app/view/__pycache__/vw_tww_wastewater_structure.cpython-311.pyc
domi4484, line 1: Should not be in git. Maybe missing gitignore?

domi4484 (approved):

ponceta:
Regarding the schemas, I'm not so concerned with their numbers as long as they remain explicit. That's maybe the main issue here which was discussed here with @rbovard: https://github.com/orgs/teksi/discussions/90#discussioncomment-7205842

If we push extensions views into the app schema it is indeed a good question of where we store additionnal tables.
A good proposal was made by @cymed about how to identify these :
https://github.com/orgs/teksi/discussions/93#discussioncomment-7324953

One of the good things with actual schemas is that you can easily drop and recreate these and define rights management on the schema level which is then inherited so as long as they remain independant from the app schema it's fine for me (which will not be the case if we include swmm views in the app schema).

The last good thing I see with dedicaced schemas is that it makes the life much easier when dealing with QGIS or FME.

ponceta:
Audit table is required for a used and documented functionnality developped by QWAT and QGEP.

https://qgep.github.io/docs/user-guide/history/history.html#history-viewer

Not ok to remove it as we use it and it can be lifesaving in certain situations.

ponceta:
OID generation is a topic we can't simply change in this development. A common system is due to any modules and the current system has already shown some limitations but it requires a good strategical decision based on what we want for the future :
I created a discussion to see what are our options :
https://github.com/orgs/teksi/discussions/94

ponceta:
I can't tell about the trigger to set the parent last modification date. For me the last modification date would be sufficient.

cymed:
>     audit table has been dropped, any objection?

See Arnaud's reply 

>    do we really need a complex oid_generation? could we live with a simple UUID? This means less maintenance, better  performance and the generation on client side is easy / possible (when editing offline for instance).

we need interlis oids.  They are required by the VSA Datamodel. For the offline editing problem, I suggest creating a table linking uuid to oid and working with uuids when using i.e. QField.

>     there is a trigger to set the parent last modification date (https://github.com/teksi/wastewater/blob/e5622cbe35f6c9a36a2fbf1d22f45bda271a3bb8/datamodel/changelogs/0001/02_sys.sql#L10C25-L10C52). Is this needed?

The last modification date is required by SIA405

>     * what PostgreSQL minimum version do we want? I would suggest 11 (https://www.postgresql.org/support/versioning/)

11 is reasonable

>     * I have added some auto formatting to the repository (pre-commit)
very nice

ponceta:
We do not support a minimal postgres version anymore. We saw that it has led users to sleep on their postgres 9 installations for too long and it has brought a lot of troubles when dealing with new functionnalities development for example.

https://github.com/teksi/Home/wiki/TEKSI-modules-releases-and-roadmap

We will support and test N-1 and actual and N+1 (For postgres it means 13-14-15 this year).

Same for QGIS by the way. 

This doesn't mean that It won't work on old db and QGIS installations. It just doesn't break or slow down our modules evolutions.

ponceta:
For aquifers and planning zones it has been proposed to integrate and document these as supported extension.

Prefix is yet to determine to theses two according again to https://github.com/orgs/teksi/discussions/93#discussioncomment-7324953

ponceta:
First test adaption is failing : (https://github.com/QGEP/datamodel/blob/master/test/test_geometry.py)

`INSERT INTO tww_app.vw_tww_reach (progression3d_geometry, rp_from_obj_id, rp_to_obj_id) VALUES (ST_SetSRID(ST_GeomFromText('COMPOUNDCURVE Z ((1 2 3,4 5 6,7 8 9))'), 2056), 'BBB 1337_0001', 'CCC 1337_0001' );`

With following error messages :

```
ERROR:  la relation « tww_sys.dictionary_od_table » n'existe pas
LINE 1: ...ELECT shortcut_en                            FROM tww_sys.di...
                                                             ^
QUERY:  SELECT shortcut_en                            FROM tww_sys.dictionary_od_table WHERE tablename = table_name
CONTEXT:  fonction PL/pgSQL tww_sys.generate_oid(text,text), ligne 22 à instruction SQL 

ERREUR:  la relation « tww_sys.dictionary_od_table » n'existe pas
État SQL : 42P01
```

ponceta:
@sjib is the new model providing both 2d and 3d geometries? Could this lead to misconfigurations / mis-use of the application?

```
ALTER TABLE tww_od.reach ADD COLUMN progression_geometry geometry('COMPOUNDCURVE', 2056);
CREATE INDEX in_tww_reach_progression_geometry ON tww_od.reach USING gist (progression_geometry );
COMMENT ON COLUMN tww_od.reach.progression_geometry IS 'Start, inflextion and endpoints of a pipe / Anfangs-, Knick- und Endpunkte der Leitung / Points de départ, intermédiaires et d’arrivée de la conduite.';
ALTER TABLE tww_od.reach ADD COLUMN progression3d_geometry geometry('COMPOUNDCURVEZ', 2056);
CREATE INDEX in_tww_reach_progression3d_geometry ON tww_od.reach USING gist (progression3d_geometry );
COMMENT ON COLUMN tww_od.reach.progression3d_geometry IS 'Start, inflextion and endpoints of a pipe (3D coordinates) / Anfangs-, Knick- und Endpunkte der Leitung (3D Koordinaten) / Points de départ, intermédiaires et d’arrivée de la conduite (coordonnées 3D)';
```

It could be problematic for some FME users since it takes always the first geometry column as default. On QGIS side its supported. We could filter the 2D column and or any undesired one with some pirogue configuration.

image

sjib:
> @sjib is the new model providing both 2d and 3d geometries? Could this lead to misconfigurations / mis-use of the application?

Up to now the VSA has a 2D model with a 3D extension to it (see https://www.vsa.ch/models)

For wastewater_structure.detailed_geometry and reach.progression_geometry I suggest to change to the 3D attributes in TEKSI wastewater (wastewater_structure.detailed_geometry3d_geometry and reach.progression3d_geometry,) as new data you collect in the field will be x,y,z. So we should provide a "container" that can store 3D data. 
If there is data without z, I think QGIS / Postgres allows to have the z-value set to NULL.

For e.g. cover we have an x,y attribute for the situation and a additional attribute with the z-level. I think this we have to keep (?)

I think we need to re-check how the current implementation works (I think with the 3D attributes), if not we should change.
And take out the 2D attributes for those two geometries in the new version.

3nids:
*schemas*

> One of the good things with actual schemas is that you can easily drop and recreate these and define rights management on the schema level which is then inherited so as long as they remain independant from the app schema it's fine for me (which will not be the case if we include swmm views in the app schema).

Views cannot be in the same schema than the tables anymore.
**So, it's either several app schemas or all in the same app schema.**
If naming is made correctly (with prefixes), we could have a way to assign roles based on the prefixes.

> The last good thing I see with dedicated schemas is that it makes the life much easier when dealing with QGIS or FME.

At seeing the number of tables per schemas (1,33 tables per schema), I don't think this is a valid argument here (even the contrary actually)

| More schemas   | Less and fixes schemas |
| ------------- | ------------- |
| + Easier roles/rights management  | + fixed and uniform across solutions  |
| (- 1 schema for 1 table is overkill)  | + easier migration path (less schemas to check, only 1 app)  |

I don't have a strong opinion here.

But just to be clear, if we stick to the current approach, we need

* tww_od
* tww_vl
* tww_sys
* tww_import
* tww_import_app
* tww_swmm
* tww_swmm_app
* tww_network
* tww_network_app
* etc.

3nids:
## audit

putting it back

## OID

So leaving as is.
@cymed if it is required to keep it, I propose to save the offline discussion for later

## min PG version

ok, I will update it in the CI

## last update

> The last modification date is required by SIA405

OK. I will move them to app then.

## Extra tables

For aquifier and planing, prefixing is a good point. But are we sure to have them in od schema?

cymed:
> > @sjib is the new model providing both 2d and 3d geometries? Could this lead to misconfigurations / mis-use of the application?
> 
> Up to now the VSA has a 2D model with a 3D extension to it (see https://www.vsa.ch/models)
> 
> For wastewater_structure.detailed_geometry and reach.progression_geometry I suggest to change to the 3D attributes in TEKSI wastewater (wastewater_structure.detailed_geometry3d_geometry and reach.progression3d_geometry,) as new data you collect in the field will be x,y,z. So we should provide a "container" that can store 3D data. If there is data without z, I think QGIS / Postgres allows to have the z-value set to NULL.
> 
> For e.g. cover we have an x,y attribute for the situation and a additional attribute with the z-level. I think this we have to keep (?)
> 
> I think we need to re-check how the current implementation works (I think with the 3D attributes), if not we should change. And take out the 2D attributes for those two geometries in the new version.

- Cover is a PointZ, Z value linked to level attribute using triggers
- reach.progression_geometry is a CompoundCurveZ. Updates on reach geometry alter the value of reach point elevation and vice versa. reach point's level is prioritised
- wastewater_structure.detail_geometry_geometry is a CurvePolygonZ already

ponceta:
> # schemas
> > One of the good things with actual schemas is that you can easily drop and recreate these and define rights management on the schema level which is then inherited so as long as they remain independant from the app schema it's fine for me (which will not be the case if we include swmm views in the app schema).
> 
> Views cannot be in the same schema than the tables anymore. **So, it's either several app schemas or all in the same app schema.** If naming is made correctly (with prefixes), we could have a way to assign roles based on the prefixes.
> 
> > The last good thing I see with dedicated schemas is that it makes the life much easier when dealing with QGIS or FME.
> 
> At seeing the number of tables per schemas (1,33 tables per schema), I don't think this is a valid argument here (even the contrary actually)
> 
> More schemas	Less and fixes schemas
> + Easier roles/rights management	+ fixed and uniform across solutions
> (- 1 schema for 1 table is overkill)	+ easier migration path (less schemas to check, only 1 app)
> I don't have a strong opinion here.
> 
> But just to be clear, if we stick to the current approach, we need
> 
> * tww_od
> * tww_vl
> * tww_sys
> * tww_import
> * tww_import_app
> * tww_swmm
> * tww_swmm_app
> * tww_network
> * tww_network_app
> * etc.

I suggest to go further with this tww_app approach and not create more than required schemas considering this principle :

EVERYTHING in the `xxx_app` schema can be dropped and reconstructed without and incidence on other schemas and tables (for the sake of the upgrade process)

ALL **ordinary tables** go to `xxx_od` and their content is editable with the user role

ALL **values lists tables** go to `xxx_vl` and their content is editable with the manager role

ALL **system tables** go to `xxx_sys` and their content is editable with the admin role

Tables linked to extensions of the base datamodel are prefixed with the extension + the extension name : ext_dss2015_aquifer, ext_dss2015_planning_zone so they are clearly identified BUT stored in dedicaced schemas  (od, vl, ...).

Additionnal Proposal IF required : we create a xxx_cfg schema for any config table that is not an od nor a vl nor a sys table (actually I just now the `tww_swmm.reach_coefficient_of_friction` in that case but there are maybe more? )

ponceta:
> > > @sjib is the new model providing both 2d and 3d geometries? Could this lead to misconfigurations / mis-use of the application?
> > 
> > 
> > Up to now the VSA has a 2D model with a 3D extension to it (see https://www.vsa.ch/models)
> > For wastewater_structure.detailed_geometry and reach.progression_geometry I suggest to change to the 3D attributes in TEKSI wastewater (wastewater_structure.detailed_geometry3d_geometry and reach.progression3d_geometry,) as new data you collect in the field will be x,y,z. So we should provide a "container" that can store 3D data. If there is data without z, I think QGIS / Postgres allows to have the z-value set to NULL.
> > For e.g. cover we have an x,y attribute for the situation and a additional attribute with the z-level. I think this we have to keep (?)
> > I think we need to re-check how the current implementation works (I think with the 3D attributes), if not we should change. And take out the 2D attributes for those two geometries in the new version.
> 
> * Cover is a PointZ, Z value linked to level attribute using triggers
> * reach.progression_geometry is a CompoundCurveZ. Updates on reach geometry alter the value of reach point elevation and vice versa. reach point's level is prioritised
> * wastewater_structure.detail_geometry_geometry is a CurvePolygonZ already

According to this statement we should have to adapt the code to use :

- situation3d_geometry instead of situation_geometry (actually missing Z dimensions for wastewater_node and cover classes on the current 2020 implementation?)
image
- progression3d_geometry instead of progression_geometry
image
- detail_geometry3d_geometry instead of detail_geometry_geometry (BUT having CurvePolygonZ )
image

**Note :** I would be in favor of commenting the 2D fields in the datamodel generation and using and displaying only 3D geometric fields. When exporting, it would be easy to force2D for the base attribute export. The naming can be the one with '3d' as the new 3D variant propose but I would avoid having 2D and 3D geometry column at the same time.

3nids:
> Tables linked to extensions of the base datamodel are prefixed with the extension + the extension name : ext_dss2015_aquifer, ext_dss2015_planning_zone so they are clearly identified BUT stored in dedicaced schemas (od, vl, ...).

> Additionnal Proposal IF required : we create a xxx_cfg schema for any config table that is not an od nor a vl nor a sys table (actually I just now the tww_swmm.reach_coefficient_of_friction in that case but there are maybe more? )

where would you put import/export/network tables? in od?

3nids:
regarding 3D, I have done this for the moment: https://github.com/teksi/wastewater/pull/4/commits/5447dd070e1c31350c55ddabc892e68088df82d2

datamodel/changelogs/0001/03_tww_db_dss.sql
```diff
@@ -1530,7 +1527,7 @@ COMMENT ON COLUMN tww_od.cover.level IS 'Height of cover / Deckelhöhe / Cote du
 COMMENT ON COLUMN tww_od.cover.material IS 'Material of cover / Deckelmaterial / Matériau du couvercle';
  ALTER TABLE tww_od.cover ADD COLUMN positional_accuracy  integer ;
 COMMENT ON COLUMN tww_od.cover.positional_accuracy IS 'Quantfication of accuarcy of position of cover (center hole) / Quantifizierung der Genauigkeit der Lage des Deckels (Pickelloch) / Plage de précision des coordonnées planimétriques du couvercle.';
-ALTER TABLE tww_od.cover ADD COLUMN situation_geometry geometry('POINT', 2056);
+ALTER TABLE tww_od.cover ADD COLUMN situation_geometry geometry('POINTZ', 2056);
```
sjib, line 1530: ALTER TABLE tww_od.cover ADD COLUMN situation3d_geometry geometry('POINTZ', 2056);

datamodel/changelogs/0001/03_tww_db_dss.sql
```diff
@@ -1171,12 +1171,9 @@ COMMENT ON COLUMN tww_od.reach.leak_protection IS 'Double-walled pipe or other s
 COMMENT ON COLUMN tww_od.reach.length_effective IS 'yyy_Tatsächliche schräge Länge (d.h. nicht in horizontale Ebene projiziert)  inklusive Kanalkrümmungen / Tatsächliche schräge Länge (d.h. nicht in horizontale Ebene projiziert)  inklusive Kanalkrümmungen / Longueur effective (non projetée) incluant les parties incurvées';
  ALTER TABLE tww_od.reach ADD COLUMN material  integer ;
 COMMENT ON COLUMN tww_od.reach.material IS 'Material of reach / pipe / Rohrmaterial / Matériau du tuyau';
-ALTER TABLE tww_od.reach ADD COLUMN progression_geometry geometry('COMPOUNDCURVE', 2056);
+ALTER TABLE tww_od.reach ADD COLUMN progression_geometry geometry('COMPOUNDCURVEZ', 2056);
```
sjib, line 1174: ALTER TABLE tww_od.reach ADD COLUMN progression3d_geometry geometry('COMPOUNDCURVEZ', 2056);

sjib: COMMENT ON COLUMN tww_od.reach.progression3d_geometry IS 'Start, inflextion and endpoints of a pipe / Anfangs-, Knick- und Endpunkte der Leitung / Points de départ, intermédiaires et d’arrivée de la conduite.';

sjib:
> Note : I would be in favor of commenting the 2D fields in the datamodel generation and using and displaying only 3D geometric fields. When exporting, it would be easy to force2D for the base attribute export. The naming can be the one with '3d' as the new 3D variant propose but I would avoid having 2D and 3D geometry column at the same time.

I would suggest to 
a) name all attributes attributename**3d**_geometry if they have a Z geometry (e.g. COMPOUNDCURVEZ or POINTZ etc. Then we know that if on the INTERLIS Import / Export the respective INTERLIS attribut does not have 3D in the name, we have to transform something.
@3nids So do not change the datatype of 2D attribute to 3D but use the 3D version and rename the attribute if you changte to Z (e.g. cover: ALTER TABLE tww_od.cover ADD COLUMN situation**3d**_geometry geometry('POINTZ', 2056);
b) I agree on commenting out the 2D attributes if there is an additional 3D attribute to not create conflicts.

ponceta:
> > Tables linked to extensions of the base datamodel are prefixed with the extension + the extension name : ext_dss2015_aquifer, ext_dss2015_planning_zone so they are clearly identified BUT stored in dedicaced schemas (od, vl, ...).
> 
> > Additionnal Proposal IF required : we create a xxx_cfg schema for any config table that is not an od nor a vl nor a sys table (actually I just now the tww_swmm.reach_coefficient_of_friction in that case but there are maybe more? )
> 
> @ponceta where would you put import/export/network tables? in od?

- [ ] qgep_import.manhole_quarantine -> tww_od.import_manhole_quarantine
- [ ] qgep_network.node -> tww_od.network_node
- [ ] qgep_network.segment -> tww_od.network_segment

(no tables for export to my knowledge)

And views will be available under tww_app.vw_export_*, tww_app.vw_network_*, tww_app.vw_import_*...

sjib:
@3nids Could you please 

- summaries the current decisions made 
- list the topics that need clarification
- draw an overview of the new (model) structure
- explain more in detail why views and data cannot be in the same schema - I do not yet understand the underlying concept for this yet.

I think this would help 
- to get an overall picture again (also in the technical group and for the jour fix next week) and 
- would allow to transfer things that are fixed into the developers guide and 
- allow to make the necessary adaptions in the datamodel creation process
- to see clearly the points that are causing bottle necks in the developing process -> they could then be addressed and decided quickly

Thanks!

datamodel/test/test_import.py
```diff
@@ -0,0 +1,476 @@
...
+            "_depth": 7.780,
+            "wn_bottom_level": 22.220,
+            "co_level": None,
+            "inlet_3_material": 5081,
+            "outlet_1_material": 5081,
+            "verified": True,
+        }
+
+        # update
+        self.update("vw_manhole", row, obj_id, "qgep_import")
```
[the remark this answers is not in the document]
ponceta: qgep_import -> tww_app

datamodel/test/test_network.py
```diff
@@ -0,0 +1,399 @@
...
+        data = {}
+        if from_id is not None:
+            data["rp_from_fk_wastewater_networkelement"] = from_id
+        if to_id is not None:
+            data["rp_to_fk_wastewater_networkelement"] = to_id
+        self.update("vw_tww_reach", data, reach_id)
+
+    def refresh_graph(self):
+        cur = self.cursor()
+        cur.execute("SELECT qgep_network.refresh_network_simple()")
```
ponceta, line 65:
```suggestion
        cur.execute("SELECT tww_app.refresh_network_simple()")
```

ponceta:
@ponceta check following value lists in new model :
image

sjib:
> @ponceta check following value lists in new model : 

maintenance_kind was moved to the subclass maintenance: 
see https://github.com/teksi/wastewater/blob/da2a86a1fd0e37c0143e953f7263ca57529d8bb6/datamodel/03_tww_db_dss.sql#L2355

and https://vsa.ch/wp-content/uploads/2023/11/wordpressexportredminehtmlblaudruck07.11.2023.html

VSAID | Betrifft | Erläuterung | Modell | Zielversion

936: de | Modellierung Erhaltungsereignisse und Subklassen - verschieben Attribut Art | Attribut Erhaltungsereignis.Art in Subklasse Unterhalt verschoben, damit Art klarer definiert ist | VSA-DSS | 2020.1 Minor release
Begründung: | In VSA-DSS ist die Klasse Erhaltungsereignisse als  ABSTRACT modelliert. Gleichzeitig ist die Art bereits in der abstrakten  Klasse. Als Konkretisierung gibt es Unterhalt und  Biol_oekol_Gesamtbeurteilung. Hier kann es Widersprüchen zwischen der  Klasse und der Art geben. Objekt aus Biol_oekol_Gesamtbeurteilung mit  Art = Sanierung_Renovierung ist möglich.

domi4484:
Tests are passing, but I had to skip some of the SWMM and all the Import as they need demo data to work.

sjib:
> Tests are passing, but I had to skip some of the SWMM and all the Import as they need demo data to work.

@domi4484 How should the demodata be provided? How did it look like for the current version (-> please provide link)

- *.backup of new database 
- *.sql dump of new database?
- other

And where should it be placed here on github?

datamodel/test/test_schemas.py
```diff
@@ -0,0 +1,40 @@
+import os
+import unittest
+
+import psycopg2
+import psycopg2.extras
+
+from .utils import DEFAULT_PG_SERVICE, DbTestBase
+
+
+class TestGeometry(unittest.TestCase, DbTestBase):
```
domi4484, line 10:
```suggestion
class TestSchemas(unittest.TestCase, DbTestBase):
```

3nids:
@ponceta this is ready for review
SQL dumps are created by the workflow: https://github.com/teksi/wastewater/actions/runs/6909987556

Triggers have been removed out of the data schemas, except for the ones in sys (generate oid + last modification) for the sake of simplicity. This should be discussed and can be change without any affectation to the QGIS project at any time.

ponceta:
Otherwise looks good to me!

ponceta:
- [ ] 2D/3D fix on wastewater_structure geometry attributes.

If I understand the logic correctly, whe should have 2d and 3d fields for detail_geometry...

```
ALTER TABLE tww_od.wastewater_structure ADD COLUMN detail_geometry_geometry geometry('CURVEPOLYGON', 2056);
...
...
ALTER TABLE tww_od.wastewater_structure ADD COLUMN detail_geometry3d_geometry geometry('CURVEPOLYGONZ', 2056);
```

Instead of :

https://github.com/teksi/wastewater/blame/12a16a3572fb551aed5e91bf8008aeaa306eb5bd/datamodel/changelogs/0001/03_tww_db_dss.sql#L379-L382

datamodel/app/view/vw_catchment_area_rwc_connenctions.sql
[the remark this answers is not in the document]
3nids: side remark: these files are never called and the views never created as far as I see.

3nids:
* Regarding wastewater structures, we have CURVEPOLYGONZ for detail_geometry_geometry and COMPOUNDCURVEZ for detail_geometry3d_geometry. Both are 3D, but one is a line, one is a polygon. @sjib what should be used here?

* Regarding commenting in the code, I would rather drop the lines which are not used. Keeping unused code makes things more confusing to me.

datamodel/app/view/vw_catchment_area_rwc_connenctions.sql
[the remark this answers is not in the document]
urskaufmann: I use these files to show the connections with lines...  happy if I have not to install them additionally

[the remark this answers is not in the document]
ponceta: @3nids can you add this to the generated app views?

sjib:
> * Regarding wastewater structures, we have CURVEPOLYGONZ for detail_geometry_geometry and COMPOUNDCURVEZ for detail_geometry3d_geometry. Both are 3D, but one is a line, one is a polygon. @sjib what should be used here?

For wastewater_structure detail_geometry3d_geometry is defined as follows in INTERLIS:

    Detailgeometrie3D: Base_LV95.Surface3D;
    Surface3D = SURFACE WITH (STRAIGHTS, ARCS) VERTEX HKoord WITHOUT OVERLAPS > 0.050;

So the equivalent in Postgres is CURVEPOLYGONZ 

Whether we want to allow ARCS I have to pass on to @ponceta @urskaufmann - do you think this is used in 3D by users?

3nids:
What I am saying is that both are named "detail_geometry(3d)_geometry" but have different geometry types (line vs polygon) and are both 3d. So there is obviously wrong either in naming or geometry type.

sjib:
> * Regarding commenting in the code, I would rather drop the lines which are not used. Keeping unused code makes things more confusing to me.

The comment would help to think about special use cases where we have to convert 3D to 2D, e.g. when exporting to the 2D version of INTERLIS models.

ponceta:
> * Regarding wastewater structures, we have CURVEPOLYGONZ for detail_geometry_geometry and COMPOUNDCURVEZ for detail_geometry3d_geometry. Both are 3D, but one is a line, one is a polygon. @sjib what should be used here?
> * Regarding commenting in the code, I would rather drop the lines which are not used. Keeping unused code makes things more confusing to me.

I do not know where this CompoundCurveZ Come from (typo?).

See my comment discussed with @sjib yesterday :

https://github.com/teksi/wastewater/pull/4#issuecomment-1819077161

- [ ] We really want these commented lines as part of the produced datamodel files to **set clear of how we differ from the original model.**

sjib:
> What I am saying is that both are named "detail_geometry(3d)_geometry" but have different geometry types (line vs polygon) and are both 3d. So there is obviously wrong either in naming or geometry type.

detail_geometry(3d)_geometry should only be used if it is a polyon
progression(3d) should be used if it is a line

3nids:
this is the code pre-refactoring: https://github.com/teksi/wastewater/blob/da2a86a1fd0e37c0143e953f7263ca57529d8bb6/datamodel/03_tww_db_dss.sql#L416-L420
there is no progression field in the code spotted. 

Regarding the comments, I can put them back.

sjib:
> there is no progression field in the code spotted.

reach has the attribute progression (to be commented out) and progression3d:
https://github.com/teksi/wastewater/blob/da2a86a1fd0e37c0143e953f7263ca57529d8bb6/datamodel/03_tww_db_dss.sql#L1212

`ALTER TABLE tww_od.reach ADD COLUMN progression_geometry geometry('COMPOUNDCURVE', 2056);
CREATE INDEX in_tww_reach_progression_geometry ON tww_od.reach USING gist (progression_geometry );
COMMENT ON COLUMN tww_od.reach.progression_geometry IS 'Start, inflextion and endpoints of a pipe / Anfangs-, Knick- und Endpunkte der Leitung / Points de départ, intermédiaires et d’arrivée de la conduite.';

ALTER TABLE tww_od.reach ADD COLUMN progression3d_geometry geometry('COMPOUNDCURVEZ', 2056);
CREATE INDEX in_tww_reach_progression3d_geometry ON tww_od.reach USING gist (progression3d_geometry );
COMMENT ON COLUMN tww_od.reach.progression3d_geometry IS 'Start, inflextion and endpoints of a pipe (3D coordinates) / Anfangs-, Knick- und Endpunkte der Leitung (3D Koordinaten) / Points de départ, intermédiaires et d’arrivée de la conduite (coordonnées 3D)';
`

sjib:
Here are the definitions for
https://postgis.net/docs/using_postgis_dbmanagement.html#CompoundCurve
https://postgis.net/docs/using_postgis_dbmanagement.html#CurvePolygon

But I cannot find CURVEPOLYGONZ - does it exist?

3nids:
> But I cannot find CURVEPOLYGONZ - does it exist?

in the link above: https://github.com/teksi/wastewater/blob/da2a86a1fd0e37c0143e953f7263ca57529d8bb6/datamodel/03_tww_db_dss.sql#L416-L420

3nids:
@ponceta I'll let you sort out the details of what you want as geometry (field name + geometry type). Let me know when you have a decision.

ponceta:
> @ponceta I'll let you sort out the details with @sjib of what you want as geometry (field name + geometry type). Let me know when you have a decision.

```
# ALTER TABLE tww_od.wastewater_structure ADD COLUMN detail_geometry_geometry geometry('CURVEPOLYGON', 2056);
# CREATE INDEX in_tww_wastewater_structure_detail_geometry_geometry ON tww_od.wastewater_structure USING gist (detail_geometry_geometry );
# COMMENT ON COLUMN tww_od.wastewater_structure.detail_geometry_geometry IS 'Detail geometry especially with special structures. For manhole usually use dimension1 and 2. Also with normed infiltratin structures.  Channels usually do not have a detail_geometry. / Detaillierte Geometrie insbesondere bei Spezialbauwerken. Für Normschächte i.d. R.  Dimension1 und 2 verwenden. Dito bei normierten Versickerungsanlagen.  Kanäle haben normalerweise keine Detailgeometrie. / Géométrie détaillée particulièrement pour un OUVRAGE_SPECIAL. Pour l’attribut CHAMBRE_STANDARD utilisez Dimension1 et 2, de même pour une INSTALLATION_INFILTRATION normée.  Les canalisations n’ont en général pas de géométrie détaillée.';
ALTER TABLE tww_od.wastewater_structure ADD COLUMN detail_geometry3d_geometry geometry('CURVEPOLYGONZ', 2056);
CREATE INDEX in_tww_wastewater_structure_detail_geometry3d_geometry ON tww_od.wastewater_structure USING gist (detail_geometry3d_geometry );
COMMENT ON COLUMN tww_od.wastewater_structure.detail_geometry3d_geometry IS 'Detail geometry (3D) especially with special structures. For manhole usually use dimension1 and 2. Also with normed infiltratin structures.  Channels usually do not have a detail_geometry. / Detaillierte Geometrie (3D) insbesondere bei Spezialbauwerken. Bei Normschächten mit Dimension1 und 2 arbeiten. Dito bei normierten Versickerungsanlagen. Kanäle haben normalerweise keine Detailgeometrie. / Géométrie détaillée (3D) particulièrement pour un OUVRAGE_SPECIAL. Pour l’attribut CHAMBRE_STANDARD utilisez Dimension1 et 2, de même pour une INSTALLATION_INFILTRATION normée.Les canalisations n’ont en général pas de géométrie détaillée.';
```

@sjib can you confirm on my syntax proposed ?

sjib:
> @sjib can you confirm on my syntax proposed ?

Ok like this!

3nids:
@ponceta I have completed all the last remarks.

ponceta (approved):

ponceta:
I go one last time through it to be sure we did not corrupt anything and then we can merge this and go forward!

ponceta:
This is good for me an closes the datamodel initial refactoring part.

Thank you all for the good work!

@3nids Ok to merge

@urskaufmann as discussed I'll provide you the .qgs project with necessary adaptations.

3nids:
@ponceta one more thought: it might be interesting to group cfg and vl schemas. I believe that permissions will be granted in a similar way and the kind of data in them are similar. If you are ok with this, I will add this too (it shouldn't affect the project as there is only one swmm table in cfg schema at the moment)

sjib:
@ponceta I saw that I did not include SRID as a variable, but fixed it on 2056. Should we change that again as before? 
`ALTER TABLE tww_od.reach_text ADD COLUMN textpos_geometry geometry('POINT', 2056);`

before in qgep
`ALTER TABLE tww_od.reach_text ADD COLUMN textpos_geometry geometry('POINT', :SRID);`

cymed:
> @ponceta I saw that I did not include SRID as a variable, but fixed it on 2056. Should we change that again as before? `ALTER TABLE tww_od.reach_text ADD COLUMN textpos_geometry geometry('POINT', 2056);`
> 
> before in qgep `ALTER TABLE tww_od.reach_text ADD COLUMN textpos_geometry geometry('POINT', :SRID);`

I would change it to :SRID. Even though we rely on the swiss standards, we shouldn't limit the usage of other EPSG. Ohterwise, it cannot be deployed i.e. humanitarian aid.

PR #25, opened 2023-11-30
remove dss2020_hydraulic_load_current from 06_swmm.sql

cymed:
use tww_od.reach.hydraulic_load_current instead


ponceta:
We have actually
`tww_od.reach.hydraulic_load_current`
`tww_od.reach.dss2020_hydraulic_load_current`
Is there a reason to have both?
If not it should be fixed in the reach class too.

cymed:
> We have actually `tww_od.reach.hydraulic_load_current` `tww_od.reach.dss2020_hydraulic_load_current` Is there a reason to have both? If not it should be fixed in the reach class too.

This PR fixes it in the reach class, see source code. TBD: adaptation in swmm views

ponceta (approved):
Small fix, inheritance of QGEP DSS 2020 fix not required anymore.

PR #29, opened 2023-12-01
restore create/drop symbology functions

### LINKED ISSUE #28: function tww_sys.drop_symbology_triggers() does not exist
SELECT tww_app.update_wastewater_node_symbology(NULL, true); calls for tww_sys.drop_symbology_triggers(), that not exist

3nids:
fixes #28


ponceta (approved):
👍

PR #36, opened 2023-12-04
Insert cover only if a cover column is not null

cymed:
At the moment, a new cover is created whenever a feature is created in vw_tww_wastewater_structure. As some ws have no cover (i.e. most discharge points and infiltration installations), we should only add a cover if any of the cover entries is distinct from NULL. 

Edit:
This PR adds a meta function check_all_nulls, which allows to check whether all entries in a jsonb starting with a specific prefix are null. By default, the postfixes `obj_id` and `identifier `are ignored. 

Example: `tww_app.check_all_nulls(to_jsonb(NEW),'co')` converts the record NEW to a jsonb. All fields starting with 'co_' are tested whether they are NULL except `co_identifier` and `co_obj_id`

The abovementioned meta function is now used to define whether a cover is added or not


cymed:
I prefer an automated solution so I can use inserts on tww_app.vw_tww_wastewater_structure in gateway exports (i.E. from the Aargau models). Otherwise keeping the gateway up to date is tedious.

cymed:
> Good idea, but I would give some user feedback, if no cover is added.

I am now raising a warning.

cymed:
requires https://github.com/opengisch/pirogue/pull/20

ponceta:
```
psql:/src/scripts/../changelogs/0001/09_tww_dictionaries.sql:122: NOTICE:  type "tww_od.plantype" does not exist, skipping
psql:/src/scripts/../app/14_geometry_functions.sql:34: NOTICE:  trigger "synchronize_level_with_altitude" for relation "tww_od.wastewater_node" does not exist, skipping
psql:/src/scripts/../app/14_geometry_functions.sql:76: NOTICE:  trigger "synchronize_level_with_altitude" for relation "tww_od.cover" does not exist, skipping
Traceback (most recent call last):
  File "/src/scripts/../app/view/create_views.py", line 273, in <module>
    create_views(
  File "/src/scripts/../app/view/create_views.py", line 193, in create_views
    vw_tww_wastewater_structure(
  File "/src/app/view/vw_tww_wastewater_structure.py", line 392, in vw_tww_wastewater_structure
    columns(
TypeError: columns() got an unexpected keyword argument 'indent'
```

3nids:
Who is reviewing this?

urskaufmann:
1. no new views pleace, because this gets also complex -> reaches must snap and get the connections
2. it must be possible to have no main cover, means to have a wastewater_structure without cover. If I delete the cover later, we have this case also...

cymed:
> What I fear is that we define default values for main_co_obj_id and main_co_identifier and this will produce a cover in almost every case and it will be difficult for the user to know how to avoid that.

We don't need to define default values for main_co_obj_id and main_co_identifier on the qgis side. The logic for the identifier default value is defined in the insert trigger. As for the obj_id, we could exclude it from the check.

>     * [ ]  Is it possible to define which classes have no cover in real life and exposes these with a view instead of creating heavy complexity?

Discharge points usually have no covers. infiltration_installations sometimes have covers and sometimes not, and there are rare cases where special structures have no covers.

ponceta:
In the actual project or even with a freshly loaded vw_tww_wastewater_structure, the co_obj_id fills itself automatically and then creates a cover.

image

image

Actually, the only workaround I know is to remove covers after wastewater structure creation.

cymed:
> Actually, the only workaround I know is to remove covers after wastewater structure creation.

I can also exempt obj_id from `{new_co_cols}` in `CASE WHEN array_length(array_remove(ARRAY[{new_co_cols}],NULL),1)>0`

datamodel/app/view/vw_tww_wastewater_structure.py
```diff
@@ -262,12 +264,25 @@ def vw_tww_wastewater_structure(srid: int, pg_service: str = None, extra_definit
...
+      CASE WHEN jsonb_strip_nulls(new_co)::text <> '' THEN -- no cover entries
+        {insert_vw_cover}
 
-      UPDATE tww_od.wastewater_structure
+       UPDATE tww_od.wastewater_structure
         SET fk_main_cover = NEW.co_obj_id
         WHERE obj_id = NEW.obj_id;
 
+     ELSE
+         RAISE WARNING 'Object_ID %: no cover created as all cover-related columns are NULL', NEW.obj_id; -- Warning
```
sjib, line 283: @cymed How does this look like in TEKSI? Can you add a screen shot?
Is 'Object_ID% a helpful text? We never used `Object_ID` anywhere; the attributes name is obj_id or co_obj_id. 
Maybe add: If you want to add a cover please fill in at least one cover attribute value.

cymed: In QGIS, the warning is shown in the log window: 
grafik
https://github.com/teksi/wastewater/pull/36/commits/47bf16b46502f1ceacd67cb12f0cf53a11b527f4 alters the warning text

cymed:
Converting to draft. Technical group should review if the warning in the log window is sufficient or whether all log window warnings should be pushed to the messageBar (must be done on the plugin side)

cymed:
@sjib was there feedback from the technical group?

sjib:
@cymed The technical group did not have time to discuss this. Can you provide a dump with this feature for easier testing?

ponceta:
- [ ] We need a test on this new use case :

1. generate a wastewater structure with no cover attributes filled 
2. expected result : the wastewater created as no)

cymed:
@sjib ready for review

cymed:
I now added a pg_notify in the trigger functions. Where would we need to add the listen/unlisten in the plugin to push the payload string to the message bar?

cymed:
> I can not check the code. My remarks to what I read in the doc: 
> - If I digitize manholes in SAA-network, I have often no cover attributes. The attribut I use most is positional_accuracy (Lagegenauigkeit). 

So we have a solution here, good.

> - If you export data for fieldwork (coordinates of manholes) it's normally the best to export the cover point. If there are manholes without cover-record, you have to be aware of this fact and use another approach.

You have to be aware anyway. A cover node does not help when there is no cover...

> - in the future I will  use positional_accuracy = unknown more conscious and more often...



We can't add a cover for every wastewater structure, some have none

cymed:
This PR does not touch the plugin. Did we introduce an error somewhere else?

sjib (approved):
Looks good for me - can we include this in a pre-release to test?

cymed:
> Looks good for me - can we include this in a pre-release to test?

From my side, yes

PR #42, opened 2023-12-05
Create vw_tww_additional_wastewater_structure.

cymed:
Tackles https://github.com/teksi/wastewater/discussions/38 
Additional ws types for drainless_toilet, wwtp_structure and small_treatment_plant are handled separately to not blow up the main ws view. Also, no extra-cols are allowed.


ponceta:
Thank you @3nids

This would be `vw_tww_additional_wastewater_structure` instead of `vw_additional_wastewater_structure`

(superclass views (superclass views that integrate all subclass attributes with prefixes to the attributes of the subclass) should be named vw_txx_superclass (example: vw_tww_maintenance_event). -> [dev_guide_views](https://github.com/teksi/Home/wiki/TEKSI-Developer-Guide#views))

cymed:
I chose to because TO is the prefix of the OID-Postfix (ch000000TO000001)

ponceta:
@cymed you can also use `toi` to avoid the postgreSQL limitation

cymed:
@sjib can we change the shortcut_en of drainless_toilet from TO to DT? DT is not used and was not used in QGEP,
Also I suggest keeping a list of all shortcuts of former tables somewhere to avoid problems with oid uniqueness.

datamodel/app/view/vw_tww_additional_ws.py
```diff
@@ -0,0 +1,576 @@
...
+            indent=6,
+            skip_columns=[],
+            remap_columns={"obj_id": "obj_id"},
+        ),
+        update_dt=update_command(
+            pg_cur=cursor,
+            table_schema="tww_od",
+            table_name="drainless_toilet",
+            table_alias="dt",
+            prefix="dt_",
```
3nids, line 521: this one should be `to` to be coherent with the OID?

cymed: prefix only I suppose?

3nids: yes, exactly

.github/workflows/datamodel-test.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: run tests
         run:  docker exec teksi-wastewater pytest
 
-      - name: failure logs
+      - name: logs
```
ponceta, line 42:
```suggestion
      - name: failure logs
```
Keep failure logs as they are only displayed if failure happens

ponceta:
On the datamodel side it seems now working but we have still to adapt something because they will appear as unkown ws_types in the current wastewater_structure layer and I think it is not expected to work like in that way (opposite is also true).

`WHERE ch.obj_id IS NULL;`

image

cymed:
> On the datamodel side it seems now working but we have still to adapt something because they will appear as unkown ws_types in the current wastewater_structure layer and I think it is not expected to work like in that way (opposite is also true).
> 
> `WHERE ch.obj_id IS NULL;`
> 
> image

As my other PR #36 also touches vw_tww_wastewater_structure I suggest I finish #36 first to not run into compatibility issues

sjib:
Are there other SQL keyword that we should avoid?
- [ ] 'AS'
- [ ] ???

sjib:
> even more ;) https://www.postgresql.org/docs/current/sql-keywords-appendix.html
OK, I checked on those also and will adapt Class shortcuts

cymed:
> On the datamodel side it seems now working but we have still to adapt something because they will appear as unkown ws_types in the current wastewater_structure layer and I think it is not expected to work like in that way (opposite is also true).
> 
> `WHERE ch.obj_id IS NULL;`
> 
> image

@ponceta Would it be beneficial to have the additional wastewater_structure types shown in the basic view, but without the updating possibilities (for hydraulics exports etc.)? i.E. 


```
        CASE
          WHEN ma.obj_id IS NOT NULL THEN 'manhole'
          WHEN ss.obj_id IS NOT NULL THEN 'special_structure'
          WHEN dp.obj_id IS NOT NULL THEN 'discharge_point'
          WHEN ii.obj_id IS NOT NULL THEN 'infiltration_installation'
          WHEN wt.obj_id IS NOT NULL THEN 'wwtp_structure'
          WHEN dt.obj_id IS NOT NULL THEN 'drainless_toilet'
          WHEN sm.obj_id IS NOT NULL THEN 'small_treatment_plant'
          ELSE 'unknown'
        END AS ws_type
[...]
WHERE ch.obj_id is NULL;
```

I don't see why the CI fails btw, as I didn't touch the insert and update scripts

datamodel/app/view/vw_tww_wastewater_structure.py
```diff
@@ -86,9 +86,12 @@ def vw_tww_wastewater_structure(srid: int, pg_service: str = None, extra_definit
...
         LEFT JOIN tww_od.wastewater_networkelement ne ON ne.obj_id = ws.fk_main_wastewater_node
         LEFT JOIN tww_od.wastewater_node wn ON wn.obj_id = ws.fk_main_wastewater_node
-        LEFT JOIN tww_od.channel ch ON ch.obj_id = ws.obj_id
         {extra_joins}
-        WHERE ch.obj_id IS NULL;
+        LEFT JOIN tww_od.channel ch ON ch.obj_id = ws.obj_id
+        LEFT JOIN tww_od.wwtp_structure wt ON wt.obj_id = ws.obj_id
+        LEFT JOIN tww_od.small_treatment_plant sm ON sm.obj_id = ws.obj_id
+        LEFT JOIN tww_od.drainless_toilet dt ON dt.obj_id = ws.obj_id
+        WHERE NULL = ALL(ARRAY[ch.obj_id,dt.obj_id,sm.obj_id,wt.obj_id]);
```
3nids: NULL is never equal to anything, you need IS keyword.

3nids (commented):
I believe this complex view requires testing, similarly to the other ones

3nids:
you apparently reused the same primary keys than for the other tests

(tests are automatically detected and run using pytest)

ponceta:
It looks technically OK for me but could you explain a bit how it is supposed to be integrated to the actual QGIS project?

Having a full `vw_tww_additional_wastewater_structure` to maintain to avoid getting a too large `vw_tww_wastewater_structure` is not clear to me in terms of advantages.

cymed:
> It looks technically OK for me but could you explain a bit how it is supposed to be integrated to the actual QGIS project?
> 
> Having a full `vw_tww_additional_wastewater_structure` to maintain to avoid getting a too large `vw_tww_wastewater_structure` is not clear to me in terms of advantages.

For now I would use `vw_tww_additional_wastewater_structure` only for import/exports for now

cymed:
even though that PR has the same % handling as vw_tww_wastewater_structure, this one throws an error

3nids:
you need to switch the double %% to simple % (next to a pirogue update)

cymed:
> you need to switch the double %% to simple % (next to a pirogue update)

then why did https://github.com/teksi/wastewater/pull/42/commits/da34826640cc81dfac0ac8a9195b9ee7f073e7a5 not work?

3nids:
because you have other remaining double %%

cymed:
@ponceta ready for review

ponceta (approved):
Let's go with this as it will soon be an issue for anyone wanting to use the new classes in TWW.

This could lead to a .qgs project adaptation and a documentation specific for these new entities.

PR #48, opened 2023-12-07
CI : Add failure logs for every failed datamodel docker CI job

ponceta:
Add failure logs for every failed datamodel docker CI job as proposed by @3nids


ponceta:
Do we want datamodel unit tests on such PR?

Issue #10, opened 2023-11-22
remark fields are limited to 80 chars


3nids:
I know that this is the VSA standard but this has proven to be an issue for users and was already discussed in QGEP.
I would recommend avoid such limitation in the data model.

ponceta:
I would propose to be flexible in the datamodel (text field) and to add a warning on the .qgs form of every attribute mentionning that when going over 80 characters, content will be truncated when exporting data to INTERLIS.

Sample possible behaviour :

1. Redlining remark field
2. Hint on what is going wrong
` length("remark") < 80`
"Attention, les remarques textuelles de plus de 80 charactères seront tronquées à l'export. Evaluez les champs appropriés pour saisir toute information pertinente."
4. Possible enforcement of the constraint on QGIS side

![image](https://github.com/teksi/wastewater/assets/8614223/6547920b-f1bd-4e62-b238-6d73ce91f290)

ponceta:
Old issue : https://github.com/QGEP/QGEP/issues/289

3nids:
Sounds like a good approach to me, let me know if I can implement this.

sjib:
The problem is that you will cut the remarks when exporting to INTERLIS and then you will loose data. You should use  the possibilites to add a file or photo if you need longer documentation for a certain object.

sjib:
Suggestion for implementation: 
Instead of 
`ALTER TABLE tww_od.data_media ADD COLUMN path  varchar(1023) ;`

New:
`ALTER TABLE tww_od.data_media ADD COLUMN path TEXT CHECK(char_length(path)<=1023)`

https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-check-constraint/
and
https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying
Would this be a good approach? @ponceta

cymed:
For those users who need the Interlis export, the limitation to 80 characters is extremely useful. Otherwise the remark field gets misused to not fill out attributes in their corresponding fields. As there is no user rights management on the QGIS side, any user can alter permissions - not good.

I like the idea of @sjib to add a check constraint on the text fields on the db side and add two functions `create_textlength_constraint `and `drop_textlength_constraint ` in `tww_sys`

3nids:
> Instead of
ALTER TABLE tww_od.data_media ADD COLUMN path  varchar(1023) ;
> New:
[...]

PR #49, opened 2023-12-07
fix re_building_group_disposal missing fk_columns

sjib:
Should fix the following issues:

- [x] https://github.com/teksi/wastewater/issues/47 - missing fk_columns
- [x] change obj_id to id as uuid - see https://github.com/teksi/wastewater/issues/47#issuecomment-1844883303 
- [x] also adapt obj_id to id as uuid for re_maintenance_event_wastewater_structure

PR #55, opened 2023-12-12
add check constraints instead of varchar

sjib:
- [x] fixes https://github.com/teksi/wastewater/issues/10
- [x] fixes tww_sys.dictionary_value_list should not be here as requested in https://github.com/teksi/wastewater/pull/49#discussion_r1423853663
- [x] [update data_media.path definitions](https://github.com/teksi/wastewater/pull/55/commits/349dc3f7492630dce6239dc0e52dbbfc3c9fe6d3)


datamodel/changelogs/0001/03_tww_db_dss.sql
```diff
@@ -61,7 +61,8 @@ WITH (
 CREATE SEQUENCE tww_od.seq_txt_symbol_oid INCREMENT 1 MINVALUE 0 MAXVALUE 999999 START 0;
  ALTER TABLE tww_od.txt_symbol ALTER COLUMN obj_id SET DEFAULT tww_sys.generate_oid('tww_od','txt_symbol');
 COMMENT ON COLUMN tww_od.txt_symbol.obj_id IS 'INTERLIS STANDARD OID (with Postfix/Präfix), see www.interlis.ch';
- ALTER TABLE tww_od.txt_symbol ADD COLUMN classname  varchar(50) ;
+ ALTER TABLE tww_od.txt_symbol ADD COLUMN classname text;
+ ALTER TABLE tww_od.txt_symbol ADD CONSTRAINT length_max_50_symbol_classname CHECK(char_length(classname)<=50);
```
3nids, line 65:
```suggestion
 ALTER TABLE tww_od.txt_symbol ADD CONSTRAINT classname_max_length CHECK(char_length(classname)<=50);
```
I would avoid extra long names (constraints are defined for tables, so the table name is implicit) and put the column identifier in front

ponceta: I do not have a strong opinion on this except the constraint name should be :

- [ ] **as understandable as possible**
- [ ] **unique** (for views like tww_ws and tww_reaches)
- [ ] **as small as possible**

sjib: I can use the shortcut of the class to still make it unique. Else we will have for each class  an `identifier_max_length`
So better `dt_identifier_max_length` (instead of drainless_toilett_identifier_max_lenght')

sjib: @3nids New version pushed - please check

3nids: They don't need to be unique as they are defined at the table level.

Again, I'd
* drop the prefix
* drop the length info (if the length changes, no need to rename the constraint)
* length_max => max_length

but it's not that big of a deal.

sjib: From a users perspective it is helpful to understand why you get the error message. Let's go ahead as technically there are no obstacles anymore.

3nids:
@ponceta you wanted to review this?

PR #58, opened 2023-12-13
update main cover if new cover is added and fk is null

cymed:
When adding a cover, if the fk_wastewater_structure has no fk_main_cover, the added cover should be used


sjib:
@cymed Would this also help to not have to run this query after an INTERLIS import?

`UPDATE qgep_od.wastewater_structure WS
SET fk_main_cover = structure_cover.co_obj_id
FROM (
  SELECT WS.obj_id AS obj_id, min(CO.obj_id) AS co_obj_id
  FROM qgep_od.wastewater_structure WS
  LEFT JOIN qgep_od.structure_part SP ON SP.fk_wastewater_structure = WS.obj_id
  LEFT JOIN qgep_od.cover CO ON SP.obj_id = CO.obj_id
  GROUP BY WS.obj_id
) AS structure_cover
WHERE structure_cover.obj_id = WS.obj_id;`

cymed:
> @cymed Would this also help to not have to run this query after an INTERLIS import?
> 
> `UPDATE qgep_od.wastewater_structure WS SET fk_main_cover = structure_cover.co_obj_id FROM ( SELECT WS.obj_id AS obj_id, min(CO.obj_id) AS co_obj_id FROM qgep_od.wastewater_structure WS LEFT JOIN qgep_od.structure_part SP ON SP.fk_wastewater_structure = WS.obj_id LEFT JOIN qgep_od.cover CO ON SP.obj_id = CO.obj_id GROUP BY WS.obj_id ) AS structure_cover WHERE structure_cover.obj_id = WS.obj_id;`

Only if the wastewater structures are imported before the covers. Otherwise just call `tww_app.update_wastewater_structure_fk_main_cover(NULL, True)` in the end.

ponceta (approved):
Looks good to me.

3nids (approved):

cymed:
PR now only adds the function tww_app.update_wastewater_structure_fk_main_cover to update_wastewater_structure_fk_main_cover. tww_app.update_wastewater_structure_fk_main_cover is defined in #97

PR #70, opened 2024-01-09
Add is_primary bool to function hierarchic vl

cymed:
Many functionalities such as SWMM or MIKE exports allow filtering elements of the primary wastewater network. Up to now, all those scripts hardcode a list of primary entries in tww_vl.channel_function_hierarchic, making it cumbersome for people that use vl extensions. To facilitate the exports (and make them quicker), I suggest adding a datamodel extensions in the vl table.


datamodel/changelogs/0001/05_data_model_extensions.sql
```diff
@@ -53,6 +53,17 @@ SET order_fct_hierarchic=
 		 ]
 	 ,code);
 
+-- this column is an extension to the VSA data model and facilitates filtering out primary features
+ALTER TABLE tww_vl.channel_function_hierarchic ADD COLUMN is_primary bool;
```
sjib, line 57: @cymed With our naming convention it should be _is_primary - isn't it?

datamodel/changelogs/0001/05_data_model_extensions.sql
```diff
@@ -53,6 +53,17 @@ SET order_fct_hierarchic=
 		 ]
 	 ,code);
 
+-- this column is an extension to the VSA data model and facilitates filtering out primary features
+ALTER TABLE tww_vl.channel_function_hierarchic ADD COLUMN _is_primary bool;
```
3nids, line 57: this needs a default value, you don't want NULL booleans

sjib:
@cymed Would also good to add the DEFAULT FALSE in the adaptions you made to tww_od.organisation see https://github.com/teksi/wastewater/pull/150/files

ALTER TABLE tww_od.organisation ADD COLUMN _active bool;
ALTER TABLE tww_od.organisation ADD COLUMN _local_extension bool;

datamodel/changelogs/0001/05_data_model_extensions.sql
cymed: or tww_is_primary (see https://github.com/teksi/wastewater/pull/164)

sjib:
Plugin test error https://github.com/teksi/wastewater/actions/runs/7898133040/job/21554937421?pr=70
should be solved with https://github.com/teksi/wastewater/pull/178 also

cymed:
Tagging @sjib for review

datamodel/changelogs/0001/05_data_model_extensions.sql
sjib: @cymed then let's do it like in #164 - then we are consistent

sjib (changes requested):
tww_is_primary instead of is_primary to be consistent with #164

sjib (approved):

PR #85, opened 2024-01-16
separate labels from ordinary data

cymed:
This pull request migrates the basic idea of https://github.com/QGEP/datamodel/pull/219 from QGEP to TEKSI Wastewater. These are the following:

- [X] add a table tww_od.tww_labels in which all labels are stored. This has the benefit that an update of a label doesn't fire any other trigger
- [X] tww_vl.channel_function_hierarchic and tww_vl.wastewater_structure_status get a column which defines whether connected channels are included in inflow/outflow labeling. 
- [X] reach points are labeled via the new function tww_app.update_reach_point_label (only sets I and O, not the value)
- [X] wastewater structure labeling is dependent on the reach point's labels
- [X] the main views need to be altered to use the new labels instead of the old
- [X] the old _label fields need to be dropped

Ideas/Comments appreciated


cymed:
As I do not have the right to set labels, please label:
datamodel
enhancement
2024.1

cymed:
@domi4484 This PR is planned to be merged in 2024.1, so I suppose we need a delta. Where do these go?

domi4484:
@cymed I don't know how it works with the delta know. let's ask @3nids 
As 2024.0 is not out yet and the release part not done maybe it make sense to include it directly in 2024.0?

cymed:
> @cymed I don't know how it works with the delta know. let's ask @3nids As 2024.0 is not out yet and the release part not done maybe it make sense to include it directly in 2024.0?

This PR requires feedback from the technical group, inclusion in 2024.0 is unrealistic

cymed:
I guess it is smarter to store the labels as a json in the first place instead of unnesting/re-aggregating. Any thoughts from the technical group @ponceta?

cymed:
Closing in favour of #409

PR #97, opened 2024-01-19
Automate updates of main cover and main wastewater node

cymed:
As discussed in TEKSI Feedbacks session from 19.01.2024. Entry point for interlis import in plugin

### DIFF
--- a/datamodel/app/symbology_functions.sql
+++ b/datamodel/app/symbology_functions.sql
@@ -104,59 +104,6 @@ $BODY$
 LANGUAGE plpgsql
 VOLATILE;
 
-  -------------------- update fk_main_cover if is null ----------------------
-
-CREATE OR REPLACE FUNCTION tww_app.update_wastewater_structure_fk_main_cover(_obj_id text, _all boolean default false)
-  RETURNS VOID AS
-  $BODY$
-  DECLARE
-  myrec record;
-
-BEGIN
-  UPDATE tww_od.wastewater_structure ws
-  SET fk_main_cover = ws_covers.co_obj_id
-  FROM (
-    SELECT ws.obj_id, min(co.obj_id) OVER (PARTITION BY ws.obj_id) AS co_obj_id
-      FROM tww_od.wastewater_structure ws
-      LEFT JOIN tww_od.structure_part sp ON sp.fk_wastewater_structure = ws.obj_id
-      LEFT JOIN tww_od.cover co ON sp.obj_id = co.obj_id
-      LEFT JOIN tww_od.channel ch ON ch.obj_id = ws.obj_id
-      WHERE (_all OR ws.obj_id = _obj_id ) AND ch.obj_id IS NULL AND ws.fk_main_cover IS NULL
-  ) ws_covers
-  WHERE ws.obj_id = ws_covers.obj_id;
-END
-
-$BODY$
-LANGUAGE plpgsql
-VOLATILE;
-
-  -------------------- update fk_main_wastewater_node if is null ----------------------
-
-CREATE OR REPLACE FUNCTION tww_app.update_wastewater_structure_fk_main_wastewater_node(_obj_id text, _all boolean default false)
-  RETURNS VOID AS
-  $BODY$
-  DECLARE
-  myrec record;
-
-BEGIN
-  UPDATE tww_od.wastewater_structure ws
-  SET fk_main_wastewater_node = ws_nodes.wn_obj_id
-  FROM (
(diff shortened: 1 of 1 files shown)


datamodel/app/symbology_functions.sql
```diff
@@ -263,6 +262,67 @@ BEGIN
...
+BEGIN
+  UPDATE tww_od.wastewater_structure ws
+  SET fk_main_cover = ws_covers.co_obj_id
+  FROM (
+    SELECT ws.obj_id, min(co.obj_id) OVER (PARTITION BY ws.obj_id) AS co_obj_id
+      FROM tww_od.wastewater_structure ws
+      LEFT JOIN tww_od.structure_part sp ON sp.fk_wastewater_structure = ws.obj_id
+      LEFT JOIN tww_od.cover co ON sp.obj_id = co.obj_id
+      LEFT JOIN tww_od.channel ch ON ch.obj_id = ws.obj_id
+      WHERE (_all OR ws.obj_id = _obj_id ) AND ch.obj_id IS NULL AND ws.fk_main_cover IS NULL
```
3nids, line 286: my question here is regarding `ws.fk_main_cover IS NULL`
if we use `_all`, don't we expect to run on all?

or can we assume, that we will never need to fix this data?

datamodel/app/symbology_functions.sql
```diff
@@ -263,6 +262,67 @@ BEGIN
 END; $BODY$
 LANGUAGE plpgsql VOLATILE;
 
+--------------------------------------------------------
+-- UPDATE wastewater structure fk_main_cover
+-- Argument:
+--  * obj_id of wastewater structure
+--  * all True to update all
+--------------------------------------------------------
+CREATE OR REPLACE FUNCTION tww_app.update_wastewater_structure_fk_main_cover(_obj_id text, _all boolean default false)
```
3nids, line 271:
```suggestion
CREATE OR REPLACE FUNCTION tww_app.wastewater_structure_update_fk_main_cover(_obj_id text, _all boolean default false)
```

datamodel/app/symbology_functions.sql
```diff
@@ -263,6 +262,67 @@ BEGIN
...
+VOLATILE;
+
+--------------------------------------------------------
+-- UPDATE wastewater structure fk_main_wastewater_node
+-- Argument:
+--  * obj_id of wastewater structure
+--  * all True to update all
+--------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION tww_app.update_wastewater_structure_fk_main_wastewater_node(_obj_id text, _all boolean default false)
```
3nids, line 302:
```suggestion
CREATE OR REPLACE FUNCTION tww_app.wastewater_structure_update_fk_main_wastewater_node(_obj_id text, _all boolean default false)
```

datamodel/app/symbology_functions.sql
cymed: Suggestion: 3 input options:

_obj_id: as is
_all : as is
_all_nulls: for automated imports

3nids: I would keep 2, but combine them

_obj_id = XXX => do XXX (don't care about NULL=
_obj_id NULL and _all is false => do only NULLs
_all is TRUE => do all

3nids: so calling without args would be the automated import

cymed: Sounds good. Do we want this setup for all the symbology functions with the arguments _obj_id and _all?

datamodel/app/symbology_functions.sql
```diff
@@ -263,6 +262,67 @@ BEGIN
...
+BEGIN
+  UPDATE tww_od.wastewater_structure ws
+  SET fk_main_cover = ws_covers.co_obj_id
+  FROM (
+    SELECT ws.obj_id, min(co.obj_id) OVER (PARTITION BY ws.obj_id) AS co_obj_id
+      FROM tww_od.wastewater_structure ws
+      LEFT JOIN tww_od.structure_part sp ON sp.fk_wastewater_structure = ws.obj_id
+      LEFT JOIN tww_od.cover co ON sp.obj_id = co.obj_id
+      LEFT JOIN tww_od.channel ch ON ch.obj_id = ws.obj_id
+      WHERE ch.obj_id IS NULL AND (_all OR ws.obj_id = _obj_id OR ( NOT _all AND _obj_id is NULL AND ws.fk_main_cover IS NULL))
```
3nids, line 286:
```suggestion
      WHERE ch.obj_id IS NULL AND (_all OR ws.obj_id = _obj_id OR ws.fk_main_cover IS NULL)
```
I think is enough

PR #110, opened 2024-01-26
fix progression_alternative

sjib:
fix progression_alternative
- [x] - rename progression_alternative to txt_progression_alternative
- [x] - add missing fk_reach to reach_progression_alternative


urskaufmann:
I hadn't realized until now that there are two *_progression_alternative tables in the database. Why two tables? reach_progression_alternative with fk_reach is enough !?
I think, it would be userfriendly, if there are some features with this layer/table. There should be a view vw_reach_progression_alternative with attributes similar to vw_tww_reach. And in vw_tww_reach there should be a calculated_attribute with info, if a progression_alternative exists for this reach (with this info the user can then configure, that the effective progression is not shown if the alternative is displayed). And there should be a function (action), that creates the alternative-geometry by moving the reach-geometry parallel.

sjib:
Originally in the development of qggep it was not clear, where we store all text and symbol labels information:

- **in one table** - therefore we had the table txt_text and txt_symbol in qgep. Consequently I have added also txt_progression_geometrie (in the first version it was named progression_geometrie)
- or **in specific tables** for each class - wastwater_structure_text, wastwater_structure_symbol, reach_text, catchment_text and following this principal now reach_progression_alternative.

So the question is, which tables to keep? I have added with x my suggestion.
- [x] reach_progression_alternative

All others we can discuss about
- [ ] wastwater_structure_text, reach_text, catchment_text - they are filled in the INTERLIS file on the export, when selecting label_export, but is this data really filled in the postgres tables also before? @domi4484 are they used on the INTERLIS export now? 
- [x] wastwater_structure_symbol - not really used until now, but should be supported in the further development of TEKSI wastewater to allow to set a position for a symbol for e.g. overview maps that is different to the position of the cover or the wastewater_node
- [ ] txt_progression_alternative - would only make sense if we would go to an approach with one central table for alternative progresssions of also other classes, not only reach.

@ponceta @urskaufmann @domi4484 @cymed Thanks for your comments

datamodel/changelogs/0001/03_tww_db_dss.sql
```diff
@@ -5318,3 +5319,5 @@ ALTER TABLE tww_od.farm ADD CONSTRAINT rel_od_farm_fk_dataprovider FOREIGN KEY (
  CREATE UNIQUE INDEX in_od_hydraulic_char_data_identifier ON tww_od.hydraulic_char_data USING btree (identifier ASC NULLS LAST, fk_dataowner ASC NULLS LAST);
  CREATE UNIQUE INDEX in_od_catchment_area_totals_identifier ON tww_od.catchment_area_totals USING btree (identifier ASC NULLS LAST, fk_dataowner ASC NULLS LAST);
  CREATE UNIQUE INDEX in_od_building_group_identifier ON tww_od.building_group USING btree (identifier ASC NULLS LAST, fk_dataowner ASC NULLS LAST);
+
+COMMIT;
```
3nids, line 5323: this line needs to be removed

sjib:
> * wastwater_structure_text, reach_text, catchment_text - they are filled in the INTERLIS file on the export, when selecting label_export, but is this data really filled in the postgres tables also before? @domi4484 are they used on the INTERLIS export now?

Do we need them if we have data in an import with imports labels (data in *_text)? How do we handle that case?

sjib:
We can also leave like this and do the discussion in a follow-up to not delay the ongoing work - I can move this topic to discussions.

sjib:
@urskaufmann Ok to merge like this? Else please comment above!

urskaufmann:
Just reach_progression_alternative. That's enough

3nids (approved):
good for me as far as I'm concerned!

datamodel/changelogs/0001/03_tww_db_dss.sql
```diff
@@ -119,29 +119,29 @@ FOR EACH ROW EXECUTE PROCEDURE
 
 -------
 -------
-CREATE TABLE tww_od.progression_alternative
+CREATE TABLE tww_od.txt_progression_alternative
```
domi4484, line 122: Is the `txt_` prefix right here? Because further down by the constraints and values is not updated

cymed:
ok for me. Handling of wastewater structures tbd in a separate PR

> * wastwater_structure_text, reach_text, catchment_text - they are filled in the INTERLIS file on the export, when selecting label_export, but is this data really filled in the postgres tables also before? @domi4484 are they used on the INTERLIS export now?

The data is directly exported from the geojson created by extractlabels_interlis. Label positions are not imported from Interlis so far.

domi4484 (commented):
This table is missing in the dictionary tables in tww_sys

sjib:
Thanks for all your feedbacks. So I see a common agreement on going for the **specific tables** approach.

So I am working on the following things:

- [ ] take out of the datamodel the following classes: txt_text, txt_symbol, txt_progression_alternative
- [ ] keeping reach_progression_alternative
- [ ] keeping wastwater_structure_text, reach_text, catchment_text in the datamodel - [but support on INTERLIS import I added to a extra issue for a next phase improvement](https://github.com/teksi/wastewater/issues/124)
- [ ] adding reach_progression_alternative to dictionary tables in tww_sys
- [ ] adding wastwater_structure_text, reach_text, catchment_text to dictionary tables in tww_sys

sjib:
@domi4484 @3nids Once I finished - Attention when merging - there are additional DEFERRABLE in https://github.com/teksi/wastewater/pull/122/files that I have not added yet in my branch

sjib:
- [ ] delete dataowner and provider in dictionary table in _text - as not existing in datamodel

sjib:
- [ ] delete dataowner and provider in dictionary table in _text - as not existing in datamodel
- [ ] rename tablename in dictionary field to re_* - else linking not working
- [ ] check attributes geometry - see https://github.com/teksi/wastewater/pull/119#issuecomment-1920801842
- [ ] check entries for extra_tables _* (currently _planning_zone and _aquifier) - see https://github.com/teksi/wastewater/pull/119#issue-2109763000

sjib:
> * check entries for extra_tables _* (currently _planning_zone and _aquifier) - see

Moved to extra issue https://github.com/teksi/wastewater/issues/134 and pull request (to be defined)

sjib:
Too many changes in the main branch - therefore replaced by https://github.com/teksi/wastewater/pull/151

PR #116, opened 2024-01-29
Alter singleinheritance order

cymed:
This PR:

- adds singleInheritance views for zone and children

- groups all singleinheritance views in a dict and creates them in a loop

- groups all plain sql scripts in 'swmm_views', 'network' and 'catchment_area' in a loop


3nids:
has this been tested with the current project?

cymed:
> has this been tested with the current project?

It is purely a db side change of how to create the views, that adds 8 additional views. All other views are not altered. New views:

- vw_tank_emptying
- vw_tank_cleaning
- vw_electric_equipment
- vw_electromechanical_equipment
- vw_solids_retention
- vw_flushing_nozzle
- vw_infiltration_zone
- vw_drainage_system

3nids (approved):
good move

PR #120, opened 2024-01-31
[WIP] Add option for role management per database

cymed:
## General
 - [ ] Fix a bug
 - [x] Add a feature
 - [ ] Maintenance / sustainability
 - [x] Add Documentation

## Describe your changes

TODO : Adapt changes definition as it now provides a structure for database, module and role naming and roles management.

In addition to the generic users ``tww_viewer``, ``tww_user``, ``tww_manager``, ``tww_sysadmin``, this PR creates database-specific pendants for cases such as https://github.com/QGEP/QGEP/issues/840

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [x] CI Tests are green
- [ ] My work is ready for review

## Checklist before merge
- [ ] Comments are resolved
- [ ] Documentation is ready


3nids:
I would keep this optional as there are good chances that there is only 1 DB instance.

Also, rather than using EXECUTE, I would opt for parameters called from psql, similarly to what we do with SRID.

cymed:
API token not found. Please provide it and it will be saved in '~/.transifexrc'.

3nids:
redgarding Transifex this should be fixed by https://github.com/teksi/wastewater/commit/35284ff86767bd51c9f9dab6398e2f4e5d30e8c7

In this PR, the new file `setup_db.sh` is overlapping with current `setup.sh`. Also, I'm not that much in favor of adding complexity to the setup until we've taken some decision on how we want to handle this.

sjib (commented):
I pass this on to @ponceta

ponceta:
https://github.com/teksi/wastewater/pull/149 has to be integrated before further here.

ponceta:
Are there any adaptations required since we will drop and recreate tww_app schema on a regular basis?

cymed:
> Are there any adaptations required since we will drop and recreate tww_app schema on a regular basis?

I will add a script to create_app.py to re-create the grants if the role exists

cymed:
ready for review @ponceta

datamodel/12_2_roles.sql
ponceta, line 1: How is {db_identifier} defined? Through SQL or python?

If it's python syntax, I'm not in favor of having .sql file with python.

CI should test this approach as well if we want to propose it to our users.

cymed: In create_app.py, db_identifier= dbname. I can use current_catalog in the plain sql script though

ponceta: Yes, I would be in favor of having plain SQL in sql files.

If you handle multiple communities and multiple networks, you will have : 
- tww_pully
- tdw_pully
- tdh_pully
- tww_belmont
- tdw_belmont
- ...

This proposal will result in having :
 - tww_viewer_tww_pully
 - tww_user_tww_pully
 - tww_manager_tww_pully
 - tww_admin_tww_pully

I suggest to have it shortened by
 - tww_viewer_pully
 - tww_user_pully
 - tww_manager_pully
 - tww_admin_pully

with less tww_ redundancy :

`SELECT concat('tww_user_'||SUBSTR(current_catalog, 5))`

And define the database naming rule for TEKSI modules in the documentation and fix the developer guide accordingly :
https://github.com/teksi/Home/wiki/TEKSI-Developer-Guide#databases-names

cymed: I will want to be able to distinguish rights for tww/tdw/tdh in the future (i.e. when working on an PGDE and a PGEE of the same community simultaneously but PGDE specialists have no PGEE knowledge)

cymed: How about SELECT concat('tww_user_'|| regexp_replace(current_catalog, 'tww_|teksi_|qgep_', ''))?

datamodel/app/tww_app_roles.sql
ponceta, line 1: @3nids .sql with python in it 
-> .sql file ?
-> .pysql file ? 
-> .py file ?

3nids: for SQL files where we only have injections, I would stick to SQL.

cymed:
Error logs do not help me in solving the documentation issue.

ponceta:
Update TX Token

3nids:
Docs building is fixed.
Regarding this PR, I still feel it needs some rework, using python and avoid EXECUTE.

cymed:
> Docs building is fixed. Regarding this PR, I still feel it needs some rework, using python and avoid EXECUTE.

I am in favour of treating all role creation scripts equally. 12_0 and 12_1 are .sql files too. 12.1 uses EXECUTE as well. Update is dynamic using python.

sjib:
@3nids Can you explain why EXCECUTE should be avoided? Is there a security issue? Or is it just a personal preference? It would be helpful to have some arguments here.

docs/en/admin-guide/security/index.rst
```diff
@@ -5,13 +5,15 @@ User Roles
...
 
-The base installation of TWW defines the following group roles :
+The base installation of TWW defines the following group roles over the entire database cluster:
 
 * **tww_viewer**: has the rights for project consultation, can select entities of TWW, view values lists.
 * **tww_user**: main TWW user, can edit entities (`tww_od` schema), view values lists.
 * **tww_manager**: user of TWW with extended privileges, can edit entities and values lists (`tww_vl` schema).
 * **tww_sysadmin**: superuser of TWW database, can edit TWW system tables (`tww_sys` schema).
 
+Their rights are cascaded to database specific group roles **tww_viewer_db_identifier** etc. if those have been set up. ``db_identifier`` is defined as ``regexp_replace(databasename, "tww_|teksi_", "")`` .
```
3nids, line 15:
```suggestion
Their rights are cascaded to database specific group roles **tww_viewer_[db_identifier]** etc. if those have been set up. ``db_identifier`` is defined as ``regexp_replace(databasename, "tww_|teksi_", "")`` .
```

3nids: regexp_replace(databasename, "tww_|teksi_", "") => this would be the default if we implement this from Python.

datamodel/12_2_roles.sql
```diff
@@ -0,0 +1,106 @@
+-------------------------------------------
+/* CREATE roles - cluster initialisation */
+-------------------------------------------
+DO $$
+DECLARE
+    role text;
+BEGIN
+    FOREACH role IN ARRAY ARRAY['tww_viewer', 'tww_user', 'tww_manager', 'tww_sysadmin'] LOOP
```
3nids, line 8: This code is duplicated with 12_0_roles.sql

docs/en/admin-guide/security/index.rst
```diff
@@ -5,13 +5,15 @@ User Roles
...
+
 Data in TWW are stored in 3 schemas with default permissions for all of these users.
 
-.. Note:: The SQL related to these roles is stored `here <https://github.com/TWW/datamodel/blob/master/12_0_roles.sql>`_ and `here <https://github.com/TWW/datamodel/blob/master/12_1_roles.sql>`_
+.. Note:: The SQL related to these roles is stored `here <https://github.com/TWW/datamodel/blob/master/12_0_roles.sql>`_, `here <https://github.com/TWW/datamodel/blob/master/12_1_roles.sql>`_ and `here <https://github.com/TWW/datamodel/blob/master/12_2_roles.sql>`_
```
3nids, line 19: After doing this review, I really believe that a proper Python command would be way clearer, nicer to use and easily share among TEKSI modules.

3nids (commented):
While all the additions seem valid, I believe this integration would deserve an implementation as a Python command for the following reasons:
* better code clarity
* easier to share among modules
* better flexibility to have roles as the DB admin wants to
* and...we already decided to have Python for app, so better use it here too.

This is a bit subjective and I don't see any technical issue at merging the PR in its current form but I doubt it is the most robust and better approach to the matter.

3nids:
also current default script (in ./datamodel/.docker/init_db.sh) creates a `tww`database. This means that this PR would create roles as `tww_viewer_tww` and similarily for user, manager and sys. This at least needs fixing.

cymed:
Errors due to missing permission in setup.sh. I cannot `chmod -x create_roles .py` from my PC (Windows) because I can't install git on the command line. Tagging @3nids

cymed:
@sjib is this PR included in TMMT?

cymed:
ModuleNotFoundError: No module named 'pgserviceparser' @3nids

datamodel/12_2_roles.sql
cymed: resolved by pythonising

cymed:
Error: SIA405_Base_Abwasser_1_LV95: model(s) not found. Why does ili2db not find https://vsa.ch/models/2020_1/SIA405_Base_Abwasser_1_2_d_LV95-20231018.ili?

sjib:
@cymed We have the 12.0_ and 12.1_roles.sql linked in the documentation: https://teksi.github.io/wastewater/installation-guide/database-initialization.html#create-minimal-roles-and-access

Shouldn't we keep those two sqls for now so that people can start also manually?
Else we should also adapt the documentation with this pull request.

sjib (changes requested):
Update documentation and describe what to do when role sql are not included anymore.

datamodel/app/create_app.py
```diff
@@ -53,13 +53,14 @@ def create_app(
...
-        "SRID": psycopg.sql.SQL(f"{srid}")
+        "SRID": psycopg.sql.SQL(f"{srid}"),
     }  # when dropping psycopg2 support, we can use the srid var directly
 
     if drop_schema:
         run_sql("DROP SCHEMA IF EXISTS tww_app CASCADE;", pg_service)
-
-    run_sql("CREATE SCHEMA tww_app;", pg_service)
+        run_sql("CREATE SCHEMA tww_app;", pg_service)
+    else:
```
3nids, line 62: why do we need to allow working on an existing schema?

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+
+from pgserviceparser import service_config
+
+try:
+    import psycopg
+except ImportError:
+    import psycopg2 as psycopg
+
+
+def get_db_identifier(pg_service: str, modelname: str):
```
3nids, line 15:
```suggestion
def get_db_identifier(pg_service: str, modulename: str):
```
don't we talk about a module rather than a model?

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+    dbname = service_config(pg_service).get("dbname")
+    db_identifier = re.sub(f"{modelname}_|teksi_", "", dbname)
```
3nids, line 17: This would require some explanations on the different possibilities and ideally a small test covering this.

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+    return db_identifier
+
+
+def create_role(pg_service: str, role: str, in_role: str = None):
```
3nids, line 21: misses docs,  especially for `in_role`

datamodel/create_roles.py
3nids, line 15: docs are missing

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+            in_role_sql = f"""
+                GRANT {in_role} TO {role};
+            """
+            cur.execute(in_role_sql)
+            print(f"granted {in_role} to role {role}")
+    conn.commit()
+    conn.close()
+
+
+def grant_privileges(pg_service: str, modelname: str, roles: list, ext_schema: str = None):
```
3nids, line 44: this misses docs

3nids, line 44: it doesn't sound super robust to have the roles as a list and use it by indexes.
I would either use a dict or specific arguments for each role.

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+    conn = psycopg.connect(f"service={pg_service}")
+    cur = conn.cursor()
+    grant_sql = f"""
+        GRANT USAGE ON SCHEMA {modelname}_od  TO {roles[0]};
```
3nids, line 48: would it be better to loop over the list of schemas to reduce the amount of code?

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+        GRANT ALL ON ALL TABLES IN SCHEMA {ext_schema} TO {roles[1]};
+        GRANT ALL ON ALL SEQUENCES IN SCHEMA {ext_schema} TO {roles[1]};
+         ALTER DEFAULT PRIVILEGES IN SCHEMA {ext_schema} GRANT ALL ON TABLES TO {roles[1]};
+        """
+        cur.execute(grant_sql)
+    conn.commit()
+    conn.close()
+
+
+def revoke_privileges(pg_service: str, modelname: str, roles: list, ext_schema: str = None):
```
3nids, line 120: same remarks than for the other method

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+    grant: bool = True,
+    ext_schema: str = None,
+    db_spec_roles: Optional[bool] = False,
+):
+    """
+    Creates the roles for usage in TEKSI Modules
+    :param pg_service: the PostgreSQL service
+    :param modelname: Abbreviation of the TEKSI model
+    :param grant: Boolean defining whether to grant privileges or revoke them
+    :param ext_schema: Schema name of the extension model
```
3nids, line 163: can you be a bit more descriptive? 
is this already fixed how this should behave?
can there be more than one extension schema?

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        cur.execute(revoke_sql)
+    conn.commit()
+    conn.close()
+
+
+def create_roles(
+    modelname: str,
+    pg_service: str = None,
+    grant: bool = True,
```
3nids, line 154: If `False` means revoking I would rather use a mode argument (probably an enum).

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+    if grant:
+        grant_privileges(pg_service, modelname, roles, ext_schema)
+    else:
+        revoke_privileges(pg_service, modelname, roles, ext_schema)
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+
+    parser.add_argument("-m", "--modelname", help="Abbreviation of the TEKSI model")
```
3nids, line 195: module?

datamodel/create_roles.py
```diff
@@ -0,0 +1,237 @@
...
+
+    parser.add_argument(
+        "-d",
+        "--database_specific_roles",
+        help="Create database specific roles",
+        default=False,
+        action=BooleanOptionalAction,
+    )
+    excl_group = parser.add_mutually_exclusive_group(required=True)
+    excl_group.add_argument(
```
3nids, line 207: after thinking a bit more, I think this script should be renamed to `manage_roles` and use subparsers. 
This is much more convenien, futureproof and a bit more natural (`manage_roles revoke` rather than `create_roles --revoke`)

You could have 3 subparsers, `create_roles`, `grant`, and `revoke` (create_roles would call grant by default).

datamodel/create_roles.py
3nids, line 120: should this be checking if the role actually exists?

3nids (commented):
This sounds good to me globally.

I added some minor remarks mainly related to code documentation and organization.

I do have one concern regarding the extensions. When discussing swmm integration in VSA-2020, it was decided not to create a dedicated schema for it (it's spread in cfg + app). I am not sure we can make such a rule.
I would either go for a more flexible API (by not talking about extension) or not integrate anything like this for now.

datamodel/create_roles.py
cymed: good point, will alter it

datamodel/app/create_app.py
cymed: I agree, not strictly necessary. Would you remove the drop_schema flag entirely?

datamodel/create_roles.py
cymed: As long as we specify the postfixes it is ok, but I wouldn't grant to all schemata in the db (e.g. pg2ili_abwasser or any end-user-specific schemata)

datamodel/app/create_app.py
3nids: I would still keep it, I found it useful when developping

cymed: We could always drop and re-create

cymed:
> @cymed We have the 12.0_ and 12.1_roles.sql linked in the documentation: https://teksi.github.io/wastewater/installation-guide/database-initialization.html#create-minimal-roles-and-access
> 
> Shouldn't we keep those two sqls for now so that people can start also manually? Else we should also adapt the documentation with this pull request.

The docs is updated with this pull request. Updating it with the last changes is pending.

cymed:
@3nids ready for re-review

docs/en/installation-guide/database-initialization.rst
```diff
@@ -71,17 +71,28 @@ There are now 6 `schemas <https://teksi.github.io/wastewater/en/user-guide/layer
...
 
-.. note:: The TWW roles are defined in the
-+ `12_0_roles.sql <https://github.com/TWW/datamodel/blob/master/12_0_roles.sql>`_ (per cluster) and
-+ `12_1_roles.sql <https://github.com/TWW/datamodel/blob/master/12_1_roles.sql>`_ (per database) files.
+- Viewer: Can consult tables and views.
+- User: Can edit data.
+- Manager: Can edit data and value lists.
+- Admin: Database administrator.
 
+.. note:: The TWW roles are defined in the python script https://github.com/TWW/datamodel/blob/master/create_roles.py. The script requires the pgserviceparser package from OpenGIS https://github.com/opengisch/pgserviceparser/.
```
ponceta, line 83: manage_roles.py

docs/en/installation-guide/database-initialization.rst
```diff
@@ -71,17 +71,28 @@ There are now 6 `schemas <https://teksi.github.io/wastewater/en/user-guide/layer
...
 
-It is recommended to use these when using TWW in a production environment.
+You can call the script from the command line with the following arguments:
+
+* ´-m´ or ´--modelname´: Abbreviation of the datamodel (here: tww)
```
ponceta, line 87: module name ?

cymed: These changes are pending, I first need the OK of the functionality before altering the docs again

ponceta:
@cymed thank you for all these changes, I think we go in the right direction with having a role management that will be replicable on all modules and flexible with multiple databases management as well.

ponceta (changes requested):
Some minor changes but otherwise we go IMO in the good direction!

docs/en/installation-guide/database-initialization.rst
```diff
@@ -167,15 +163,5 @@ You can also generate the data model under Linux.
...
-If you already have a data model and you want to force the regeneration
-of the model you can also use the ``-f`` option: ``./scripts/db_setup.sh -f``.
-
-You can use the ``-r`` option to add roles (``tww_viewer``, ``tww_user``, ``tww_manager``, ``tww_sysadmin``).
-
-- Viewer: Can consult tables and views.
-- User: Can edit data.
-- Manager: Can edit data and value lists.
-- Admin: Database administrator.
+If you want to use a different SRID, alter the SRID definition.
```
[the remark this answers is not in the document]
cymed: it can be set in setup.sh as an ENV. This PR does not alter SRID handling

ponceta:
This is great, the last thing we need is a way to launch these python functions from TMMT or temporary from the plugin.

Or let the SQL files temporary available, because many people are lost when doing python from the command line.

One huge actual limitation is that we actually can't create a database with a pg_service connection. Therefore you have to go to pgAdmin, create a database, then go to python or QGIS to initialise your roles and then go back to pgAdmin to restore the datamodel dump.

cymed:
I now added a Plugin menu item to create database roles. It is only active if the admin mode is enabled and checks for superuser rights before doing anything

sjib:
@ponceta Is this ready for 2024.0 and can be merged?

ponceta:
This will remove  the SQL script with pgAdmin for rights management, I think wait for 2024.1 or 2025.0 is the best way to go. This is strongly linked to TMMT.  (Same as https://github.com/teksi/wastewater/pull/399)

cymed:
Converting to draft, check with TMMT how to proceed

cymed:
Superseded by #625. Role Management needs to be done inside PUM

PR #128, opened 2024-02-01
add drop_views function

cymed:
in create_views, we used to have a drop_views function. Imho it is smart to warn the user before dropping, the final functionality is tbd though.

### DIFF
--- a/datamodel/app/view/create_views.py
+++ b/datamodel/app/view/create_views.py
@@ -20,33 +20,11 @@ def run_sql(file_path: str, pg_service: str, variables: dict = {}):
     conn.close()
 
 
-def create_views(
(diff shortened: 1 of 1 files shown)


ponceta:
👍 We will not implement a drop view logic if it can be avoided!

cymed:
OK. Where is the app schema dropped and re-created in the code? I'd like to set up my extensions to re-create all views after running an init script

3nids:
This is not yet done (would be part of a follow-up project).

but basically there are the lines to be run at the moment: https://github.com/teksi/wastewater/blob/d346f4de0e69c023ae9105f897a701c33a71636d/datamodel/scripts/setup.sh#L33-L39

PR #142, opened 2024-02-02
[WIP] Add Tool to split reaches with wastewater nodes and points

cymed:
This WIP will eventually add a tool to create a wastewater structure or a wastewater node on a reach and split the reach in two parts. The Reaches' corresponding Wastewater Structure and Channel attributes are preserved for wastewater node but re-created for Wastewater structure. The old reach is dropped.

A new button is added
grafik
that adds a popup in analogy to the wizard
grafik
and then lets you digitise a wastewater structure or wastewater node. The reach that is snapped upon is split by the node (there is a bug in that step for now, see  below)

TBD:
Programming:
- [] The code tries to use the vertex Id to create two new reaches. However, `QgsGeometry.vertexIdFromVertexNr(vertexIndex)` seems to have problems when `vertexIndex==0`.  The same problem exists for TwwMapToolAddReach. Due to these problems, the current code cannot be fully tested

Usability (Tagging @urskaufmann and @tomep for input)
- [ ] When splitting a reach by a wastewater Structure, two new channels are created and the old reach is dropped. How do we want to proceed for easy KEK compatibility and updates of Catchment areas/connection points referencing the old reach?
- [ ] Which fields should be kept from the previous vw_tww_reach when splitting with wn/ws?
- [ ] the default elevations of (wn_)bottom_level are extracted from the reach geometry. Is this ok? Reaches with more than two vertices that were digitalised in 2D might have NULL or  0 as vertex elevation 

Note: This PR needs thorough testing and will not make it before the feature freeze. Aiming for 2024.1


urskaufmann:
when splitting a reach the user must have this options:
- split channel also or just split reach
- add a new wastewater-node or not
- manually edit the distance, where the split is (default: distance of the user-click)
- manually edit the reachpoint-level at the split-point (default: is calculated average at the splitpoint, if there are levels at the old reachpoints)

The tool has then to:
Any case:
-	Create new reach and copy all attributes reach/wastewater-networkelement/reachpoint except OBJ_ID and IDENTIFIER
-	Split geometry at splitpoint and delete lower part (old reach) or upper part (new reach)
-	Set old to-reachpoint level to splitlevel, set new from-reachpoint level to splitlevel
-	Control if reach-to-reach-connections (Blindanschlüsse) exist to the old reach
-	If exist reach to reach connections: message to the user to control the connections. Or better: control which connections are to the new reach and correct them. But this may be not always possible without doubt, if a connecting reach is not drawn exactly to the old/new reach

If new wastewater_node:
-	Create new wastewater_node and wastewater_networkelement
-	To discuss: Set fk_wastewater_structure to the old (reach)wastewater_networkelement.fk_wastewater_structure or let it NULL
-	Set old to-reachpoint.fk_wastewater_networkelement to the OBJ_ID of the new wastewater_node
-	Set new from-reachpoint.fk_wastewater_networkelement to the OBJ_ID of the new wastewater_node
-	Set wastewater_node.bottom_level to the splitlevel

If no wastewater_node:
-	Set old to-reachpoint.fk_wastewater_networkelement to the OBJ_ID of the new reach
-	Set new from-reachpoint.fk_wastewater_networkelement to NULL

If split channel:
-	Create new channel/wastewater_structure and copy all attributes channel/wastewater_structure from old channel except OBJ_ID and IDENTIFIER
-	Set new-reach/wastewater_networkelement.fk_wastewater_structure to the OBJ_ID of the new channel
-	Duplicate all records in re_maintenance_event_wastewater_structure for the old channel and add them for the new channel. 

-	The user has to control, if the related maintenance_events are correct for old and new channel

There are never automatic changes to connections of catchment_areas (also because catchment_areas are normally not connected to reaches).

urskaufmann:
Split without node: in secondary network is no need of nodes for reach to reach connections.

Channel is not defined from manhole to manhole, if there is another primary channel connected in between without manhole.

manually splitlevels are needed if the slope is not continues (e.g. you have measured the channel at this point). Otherwise you have to correct it manually afterwoods 2 or 3 times (if you don't forget it).

length effective is calculated 

I want all attributes to be copied. The user has then normally to change at least one attribute (e.g. material and he should then think about the other attributes, that a PP channel will not have the same construction_year as the normal_concrete channel).

If there are at the end two new reaches or an old and a new one does not matter. Do describe it's easier to take about old and new...

urskaufmann:
Annotation 2024-03-19 145329
Example: my old userinterface: I have a button in the taskbar, click on a reach and get this window. The distance of the click is prefilled in the window, also the level of the reach at this point (in the example 19.5m from the manhole above). I can change the distance and the level and choose, if channels are "split" = duplicated and if a node is set. If I want to place a manhole, I do this afterwards manually (with wizard)

sjib:
`error: cannot format plugin/teksi_wastewater/tools/twwmaptooladdfeature.py: Cannot parse: 767:39:             assert len(new_geometries) = 2`

cymed:
Closing in favour of a pending draft pr

PR #150, opened 2024-02-07
add _active to tww_od.organisation

cymed:
this PR adds a column `_active` to `tww_od.organisation` to facilitate filtering the loaded organisations


cymed:
added a _local_organisation bool that can be used in the future for the interlis export. Plugin functionality not part of this pull request, my suggestion is to tackle it in part 2 of VSA DSS development

sjib (approved):
Good improvment and preparation for the management of organisations with VSA-DSS 2020

PR #153, opened 2024-02-08
DSS 2020 export hotfix

cymed:
In the exporter to intermediate schema, some _export_foobar functions made their queries to control_center instead of the class itself. This hotfix should fix the problem for unfiltered layers, I do not have the time to set up the query filters


domi4484 (approved):

cymed:
@domi4484 check fails due to some transifex token. Merge anyway?

domi4484:
> @domi4484 check fails due to some transifex token. Merge anyway?

Seems a token for transifex is missing:
```
qgis-plugin-ci package: error: argument --transifex-token: expected one argument
```

@3nids any idea? should the token be in the secrets?

3nids:
test fixed in bd5c5693add367dd36e446c652ca75d776870944

PR #155, opened 2024-02-08
Enable cmd tool tests

### LINKED ISSUE #138: enable command line testing in workflow

domi4484:
Fix #138


3nids:
it's strange that the service is not working

3nids (approved):

domi4484:
> it's strange that the service is not working

I think the interlis tools are not able to use it

3nids:
let's see in #157

domi4484:
If with service file doesn't work we should remove the --pgservice option from the cmd line tool. Or raise a not implemented error.
In model baker we had to implement something to parse the pg service and pass the right options to the ili tools. And here we don't use (that part) of MB.

Issue #107, opened 2024-01-26
Datamodel - Cyclic relations on classes missing


sjib:
The attribute and relation log_card.fk_main_structure and fk_next_special_building are missing in the datamodel.

!!@ comment = "Assoziation"
ASSOCIATION Stammkarte_HauptbauwerkAssoc =
  HauptbauwerkRef  -- {0..1} Stammkarte;
  Stammkarte_HauptbauwerkAssocRef -- {0..*} Stammkarte;
END Stammkarte_HauptbauwerkAssoc;

!!@ comment = "Assoziation"
ASSOCIATION Stammkarte_Naechstes_SBWAssoc =
  Naechstes_SBWRef  -- {0..1} Stammkarte;
  Stammkarte_Naechstes_SBWAssocRef -- {0..*} Stammkarte;
END Stammkarte_Naechstes_SBWAssoc;

sjib:
Same with measuring_point.fk_reference_station

 !!@ comment = "Assoziation"
ASSOCIATION Messstelle_ReferenzstelleAssoc =
  ReferenzstelleRef  -- {0..*} Messstelle;
  Messstelle_ReferenzstelleAssocRef -- {0..*} Messstelle;
END Messstelle_ReferenzstelleAssoc;

sjib:
These are all relations on the class itself.
@ponceta @domi4484 Can they be treated like other relations or should there be an extra table xxx_hierachy?

Many years ago I implemented this on MS ACCESS like this. Is this still the way to go also with Postgres?

```
                'RE_MESSSTELLE_HIERARCHIE
                Print #1, "CREATE TABLE RE_MESSSTELLE_HIERARCHIE( OBJ_ID  Text(" & OBJ_IDLengthString & ") not null, FS_MESSSTELLE Text (" & OBJ_IDLengthString & "), FS_REFERENZSTELLE Text (" & OBJ_IDLengthString & "));"
                
                Print #1, "ALTER TABLE RE_MESSSTELLE_HIERARCHIE ADD CONSTRAINT medrel204 FOREIGN KEY (FS_MESSSTELLE) REFERENCES MESSSTELLE(OBJ_ID);"
                Print #1, "ALTER TABLE RE_MESSSTELLE_HIERARCHIE ADD CONSTRAINT medrel205 FOREIGN KEY (FS_REFERENZSTELLE) REFERENCES MESSSTELLE(OBJ_ID);"
 ```

sjib:
[...]

PR #156, opened 2024-02-09
add dictionary sql for extra dss2020 extension classes

sjib:
- [x] 51_aquifier_2015_dictionaries.sql
- [x] 52_planning_zone_2015_dictionaries.sql
- [x] adapt script setup.sh

PR #158, opened 2024-02-10
add missing cyclic relations

sjib:
should solve https://github.com/teksi/wastewater/issues/107

### DIFF
--- a/datamodel/changelogs/0001/03_tww_db_dss.sql
+++ b/datamodel/changelogs/0001/03_tww_db_dss.sql
@@ -1,6 +1,6 @@
 ------ This file generates the VSA-DSS database (Modul VSA-DSS (2020)) in en on QQIS
 ------ For questions etc. please contact Stefan Burckhardt stefan.burckhardt@sjib.ch
------- version 07.02.2024 20:37:04
+------ version 10.02.2024 11:44:42
 ------ with 3D coordinates
 
 ---------------------------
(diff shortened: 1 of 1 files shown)


sjib:
@3nids Are these the right options for a cyclic relation?
` ON UPDATE CASCADE ON DELETE SET NULL DEFERRABLE INITIALLY DEFERRED`

ponceta:
@ponceta check the defferable setting with @3nids before merging this.

3nids (approved):

ponceta:
The differences between DEFERRABLE INITIALLY DEFERRED, DEFERRABLE INITIALLY IMMEDIATE and NOT DEFERRABLE are as follows:

* DEFERRABLE INITIALLY DEFERRED means that the constraint is deferred until the end of the transaction, unless the SET CONSTRAINTS command is used to make it immediate.
* DEFERRABLE INITIALLY IMMEDIATE means that the constraint is checked at the end of each instruction, unless the SET CONSTRAINTS command is used to make it deferred.
* NOT DEFERRABLE means that the constraint is always checked at the end of each instruction, and that the SET CONSTRAINTS command cannot be used to modify it.

In summary, DEFERRABLE and NOT DEFERRABLE indicate whether the constraint can be deferred or not, and INITIALLY DEFERRED and INITIALLY IMMEDIATE indicate the default checking mode for the constraint.

@sjib, @3nids, @cymed, @domi4484  for TEKSI would you be OK to use the DEFERRABLE INITIALLY DEFERRED for every relation connecting objects with each other in the database? (at least where it makes sense, so we can set it once and for all for the database)

sjib:
> @ponceta Good for me - please merge this and #156 I will create a version with DEFERRABLE INITIALLY DEFERRED in a new pull request based on the updated master.

@ponceta Here the additional pull request: https://github.com/teksi/wastewater/pull/178

PR #161, opened 2024-02-13
add vw_farm

cymed:
Adds a SingleInheritance View on Farm - Building Group


cymed:
@sjib fk_dataowner and fk_provider on tww_od.farm are duplicates of tww_od.building groups. Shall I adapt the model in this PR or do you want to do it in the metadata?

sjib:
@cymed So you also have to rethink your view!

sjib:
For reference: Thoughts about default dataowner / provider see https://github.com/teksi/wastewater/issues/135

PR #164, opened 2024-02-14
Add  DEFAULT FALSE on organisation flags

cymed:


datamodel/changelogs/0001/05_data_model_extensions.sql
```diff
@@ -72,12 +72,12 @@ COMMENT ON COLUMN tww_od.wastewater_node._status IS 'not part of the VSA-DSS dat
 added solely for TEKSI Wastewater & GEP
 has to be updated by triggers';
 
-ALTER TABLE tww_od.organisation ADD COLUMN _active bool;
+ALTER TABLE tww_od.organisation ADD COLUMN _active bool DEFAULT FALSE;
```
3nids, line 75: should be true, no?
I would expect an organisation added to be active by default. Or am I missing something?

sjib: - When we import the VSA organisation dataset with 2100 organisation - do we want all organisations to be active or not? I would say rather not - I just want to set those active that I need afterwards - that was my argument to set the DEFAULT FALSE
- When I create a new organisation in TEKSI , I can set the _active flag myself to TRUE

cymed: Calling
```
UPDATE tww_od.organisation
SET _active=TRUE
WHERE obj_id = ANY(
SELECT UNNEST(ARRAY[fk_provider,fk_dataowner])
FROM tww_od.wastewater_networkelement
UNION
SELECT UNNEST(ARRAY[fk_owner,fk_operator])
FROM tww_od.wastewater_structure
UNION
SELECT fk_operating_company
FROM tww_od.maintenance_event
)
```
after import would set the active flag correctly 99% of the time after import. @sjib do you see other classes we should check for in such a query?

3nids: I think the logic should be when using the application and not when running import where we can easily play with the data. The defaults should be the ones you need when using QGIS.
That's how I see it.

sjib: @cymed 
> after import would set the active flag correctly 99% of the time after import. @sjib do you see other classes we should check for in such a query?

We have also 
Massnahme.TragerschaftRef
Masnnahme.Verantwortlich_AusloesungRef
Stammkarte.BueroRef
Stammkarte.StandortgemeindeRef
Messstelle.BetreiberRef

sjib: Good thoughts from different angles. 
I see three use cases with organisations :

1. Importing organisation datasets ahead of datasets via INTERLIS import
2. Importing new VSA organisations that we have applied for or additional that we need as for example our TV company partner or summing organisation where we match local organisations to
3. Creating new (local) organisations during operation in TEKSI (should be avoided if possible, as the goal is to integrate all organisations used to the VSA organisation dataset).

Are there others?

ponceta:
`_active` should be reserved for calculated field where the value is computed, `active` or `tww_active` would be preferable.

It is a core feature of teksi modules, I would be in favor of having it as simple as possible since it will be replicated to all modules.

sjib (approved):

PR #166, opened 2024-02-15
include overflows in symbology

cymed:
There are several situations (i.E. a stormwater composite tank with a preceding separating structure) where a wastewater structure only has overflows as topological connections. In these situations, we should also set a _usage_current dependent on the incoming overflow. This PR includes them in the updating process for wastewater nodes.

See also https://github.com/QGEP/datamodel/pull/225/files


cymed:
@ponceta ready for review

ponceta:
It looks good, does it works good too?

Have you tried it on with a data set? My main concern is the performance, if it's green let's go!

cymed:
> It looks good, does it works good too?
> 
> Have you tried it on with a data set? My main concern is the performance, if it's green let's go!

It works with real data. 
For performance, I have tested the `SELECT DISTINCT ON ...` bit with _all=true on a small db of 2'570 wastewater nodes. 

Before: 
Planning Time: 3.110   ms
Execution Time: 77.668 ms

New: 
Planning Time:   12.957 ms
Execution Time: 129.622 ms

For a single wastewater node (i.e. when altering a wastewater structure)
Before: 
Planning Time: 3.188   ms
Execution Time: 0.216 ms

New: 
Planning Time:   8.382 ms
Execution Time: 0.525 ms

If the performance drop bothers you I suggest to store` _usage_current`, `_status `and `_function_hierarchic ` on tww.overflow and grab it from there

ponceta (approved):

ponceta:
I'm OK to go with that but I would enjoy a good schema in the doc on how it works and what special structures can be well modelized with that, otherwise it is quite an additionnal complexity in terms of symbology.

I think it is mainly on the overflows "new" mechanics and it would be good if one can achieve it without being a wastewater datamanager certified engineer. (otherwise I'm quite excited to see how it looks in real life examples.)

cymed:
> I'm OK to go with that but I would enjoy a good schema in the doc on how it works and what special structures can be well modelized with that, otherwise it is quite an additionnal complexity in terms of symbology.
> 
> I think it is mainly on the overflows "new" mechanics and it would be good if one can achieve it without being a wastewater datamanager certified engineer. (otherwise I'm quite excited to see how it looks in real life examples.)

I would not add any additional docs, as the correct topology is defined in the vsa wiki. It is used for cases such as this
grafik
We have an overflow flowing into a cso tank. the tank has an overflow into the water body. Here, the cso tank's wastewater_node did not receive a _usage_current / _status /_function_hierarchic, as there was no topological connection to a reach.  With the PR, the result is

The last pushes improve general performance as there are fewer bulky left joins.

cymed:
The last pushes improve performance and readability. The only change to the original function is that I included the wn's corresponding ws for status.

All the overflow specific stuff is separated from the other PR, making it much faster as it left joins the nodes onto an overflow instead of left joining the overflow on the node.

> I'm OK to go with that but I would enjoy a good schema in the doc on how it works and what special structures can be well modelized with that, otherwise it is quite an additionnal complexity in terms of symbology.
> 
> I think it is mainly on the overflows "new" mechanics and it would be good if one can achieve it without being a wastewater datamanager certified engineer. (otherwise I'm quite excited to see how it looks in real life examples.)

I don't think we need another docs. It only affects symbology on wastewater nodes for those who follow the vsa wiki. The rest won't notice.

Here is a real-life example, where there is an separating structure is connected to a combined sewer overflow with retention volume , which in turn has an overflow into the water body. Without the PF, the CSO has no usage_current
grafik

sjib:
@cymed @ponceta  May be we could include this screenshot in this chapter: https://qgep.github.io/docs/user-guide/How-To/index.html#hydraulic-modeling-of-an-overflow-prank-weir-leapingweir-pump

PR #170, opened 2024-02-22
fix reach direction change

cymed:
Reach direction change was falsely connected to `tww_od.reach_direction_change(text[])` instead of `tww_app.reach_direction_change(text[])`


domi4484 (approved):

3nids:
we can safely merge this, the bug is coming from a plugin-ci issue (fixed here https://github.com/opengisch/qgis-plugin-ci/pull/287, release will come a bit later)

PR #171, opened 2024-02-22
Dss export: fix organisation.bemerkung length and allow vl extensions

cymed:
changed from 80 to 255, see https://vsa.ch/models/2020_1/SIA405_Base_Abwasser_1_2_d_LV95-20231018.ili Line 139


plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py
```diff
@@ -445,7 +447,7 @@ def _export_organisation(self):
                 **self.sia_405_base_common(row, "organisation"),
                 # --- organisation ---
                 auid=row.uid,
-                bemerkung=self.truncate(self.emptystr_to_null(row.remark), 80),
+                bemerkung=self.truncate(self.emptystr_to_null(row.remark), 255),
```
sjib, line 450: As we have now the constraints in the datemodel

https://github.com/teksi/wastewater/blob/42cbbec329a57f5e436c1328d2359c0261fb3763/datamodel/changelogs/0001/03_tww_db_dss.sql#L172

I am not sure if we still need the truncate here. If yes the correction is correct.

sjib:
Another exception to length of remark is with abwasserreinigung (also 255 instead of 80):

https://github.com/teksi/wastewater/blob/a427c63ec84ecf308051328f3609049229a80a0e/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py#L989

sjib (approved):
Check comment on truncate, else it looks good (without having tested if the use_vsa is working - but the option to use vsa code is welcomed.

cymed:
Error is: 

```
=================================== FAILURES ===================================
_________________ TestInterlis.test_dss_dataset_import_export __________________
[...]
psycopg2.errors.ForeignKeyViolation: insert or update on table "mechanical_pretreatment" violates foreign key constraint "rel_mechanical_pretreatment_wastewater_structure"
DETAIL:  Key (fk_wastewater_structure)=(ch080qwzVA001062) is not present in table "wastewater_structure"
```
Where is the script of TestInterlis.test_dss_dataset_import_export? I am not sure that the unit test fails because of the alterations

PR #173, opened 2024-02-26
Add default values for fk_provider and fk_dataowner

cymed:
This pr adds a metatable in tww_od (in first few pushes it was tww_sys) and the corresponding function to manage default values of any obj_id type. For now, only fk_provider and fk_dataowner are implemented. The functionality can easily be expanded for any obj_id based fk i.e. fk_pipe_profile.


datamodel/changelogs/0001/14_default_values.sql
urskaufmann, line 1: good idea to define this defaults. Is it correct to have this new table default_values in sys schema? Or better in app schema?
These default_values will be added when inserting. And also with update of any other attribute of the record and overwrite existing data in fk_dataowner and fk_provider? That would not be a good idea.
Mentioning the OID prefix in line 15 seems to be copied from another sql and has no meaning in this sql

cymed: app is re-created on every update. Either sys or od. default values are only used on create statements.

sjib:
Should solve https://github.com/teksi/wastewater/issues/135

sjib:
@cymed Can you add some documentation how to use this feature? Where do you add or select the default dataowner / provider? Screenshots?

cymed:
Is the solution ok like this (default handling by sysadmin only in tww_sys)?

sjib:
> Is the solution ok like this (default handling by sysadmin only in tww_sys)?

@urskaufmann @ponceta @varrieta and others - your more in daily business

urskaufmann:
In most cases, this should be ok. I have some databases with more than one dataowner. In this case it would be better, when also non-admin-users can change the default value (or I show them how to set a default in the attributform. If I set a default in the attributform, will it overwrite the default, that comes from tww_sys?)

cymed:
> In most cases, this should be ok. I have some databases with more than one dataowner. In this case it would be better, when also non-admin-users can change the default value (or I show them how to set a default in the attributform. If I set a default in the attributform, will it overwrite the default, that comes from tww_sys?)

if you want regular users to alter the default values, it must not be in tww_sys as tww_user does not have the necessary rights for it.

cymed:
@ponceta ready for review, feedback from @urskaufmann was included in [#615edd6](https://github.com/teksi/wastewater/pull/173/commits/615edd66ef3ca1cbe910a12d780ed94d20056f00)

docs/en/user-guide/How-To/index.rst
```diff
@@ -216,6 +216,17 @@ The Feature Attributes window for the overflow characteristic appears:
...
 3. Go back in the field fk_overflow characteristic of the just created overflow characteristic with the button **switch to form view** and define the necessary HQ or QQ values with the **add child object** button.
 
+Setting default values
+-----------------------
+
+It is possible to define default values for repeated fields such as fk_provider or fk_dataowner.
+
+1. Enter field name and value in the corresponding attributes in ``tww_od.default values``
+
+2. Connect all ccurences of your field name to ``tww_sys.get_default_values(field_name)``.
```
ponceta, line 226:
```suggestion
2. Connect all occurences of your field name to ``tww_sys.get_default_values(field_name)``.
```

datamodel/changelogs/0001/14_default_values.sql
```diff
@@ -0,0 +1,50 @@
+CREATE TABLE tww_od.default_values
```
ponceta, line 1: You mention a table in tww_sys in your PR description and it is stored in tww_od in your code.
-> tww_od : any user can edit these values
-> tww_vl : a manager user can edit these values
-> tww_sys : an administrator user can edit these values
- [ ] Depending on who will be able to define these default values, the good schema has to be chosen.

cymed: I moved it from sys to od because @urskaufmann has use cases where a tww_user needs the right to alter the values

ponceta:
This is great and simplify greatly the fk_provider and fk_dataowner process.

ponceta (approved):

PR #174, opened 2024-02-27
Add view for detailgeometry

cymed:
With #68, the fields _usage_current and _function_hierarchic were dropped from tww_od,wastewater_structure. In order to access these informations, we need a view with a left join on wastewater node to be able to use _usage_current and _function_hierarchic in visualisations in QGIS (i.e scale dependability, color etc.) This PR adds a new view tww_app.vw_wastewater_structure for this very purpose.


ponceta:
Looks good, have you tested it on data / QGIS project?

cymed:
I have tested using it in visualisations, but I have not replaced the existing layer nor tested the digitising actions.

ponceta (approved):
This is the database side of the new tww_app.wastewater_structure, meant to replace actual tww_od.wastewater_structure.

Works exactly the same except it has inheritance on _channel_function_hierarchic and _channel_usage_current.

We will be able to define main node and main cover from this layer too if required.

Project adaptations will be required to do so.

PR #175, opened 2024-03-01
add function to calculate catchment area totals from catchment areas

cymed:
This PR adds a function in analogy to the symbology functions to calculate the catchment area totals from the catchment area entries. It only does so for the catchment_area_totals linked to a main log card Aggregation follows the following path

Catchment -> log card -> main log card -> pwwf_wastewater_node <- hydr_char_data  <- catchment_area_totals

So far, it is not implemented as a trigger function, but to be called manually


cymed:
Currently, it does not aggregate the child records of the log card, but only the directly connected areas. We need some kind of recursive CTE to calculate the aggregates

sjib:
@cymed I tried to test this. Can you add an example sql statement to run this manually here?

cymed:
> @cymed I tried to test this. Can you add an example sql statement to run this manually here?

first call the sql statement using copy-paste, then run the function using `SELECT tww_app.update_catchment_area_totals(NULL, true)` to update all.

sjib:
@cymed And for a specific obj_id of catchment_area_totals: `SELECT tww_app.update_catchment_area_totals('ch123456CM000001, false)` ?

cymed:
false is default and can be omitted

sjib:
`SELECT tww_app.update_catchment_area_totals('ch123456CM000001')`

ponceta (approved):

PR #179, opened 2024-03-05
add view for catchment area totals

cymed:
This adds an aggregate view of the catchment area totals and the corresponding main wastewater node for gep visualisation


urskaufmann:
to use or test this, there has to be configurations in catchment_area and log_cards and in best case also good tools to automatic set records and values...
not able to test actually

ponceta (approved):

PR #185, opened 2024-03-13
Upgrade Postgis

### LINKED ISSUE #20: Current code is invalid with Postgis 3.2+
We currently use null values for Z coordinates in linestrings using NaN.
For instance in reach insert trigger: https://github.com/teksi/wastewater/blob/fbc9c16cc5f1b8a078429b04779601feb050e8f1/datamodel/app/view/vw_tww_reach.py#L176-L180

Starting with Postgis 3.2, this fails with` Geometry contains invalid coordinates` error message.

I don't know how to handle this yet and have posted an issue on stackexchange and osgeo trac.

SE: https://gis.stackexchange.com/questions/470332/how-to-handle-null-z-m-values-in-postgis
osgeo ticket: https://trac.osgeo.org/postgis/ticket/5610#ticket

3nids:
fixes #20 

that would probably deserve some documentation about invalid PG versions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,8 @@ TWW_PG_PORT=5432
 
 QGIS_TEST_VERSION
 
+POSTGIS_IMAGE=postgis/postgis:14-3.4
+
 # fix access rights for schemaspy
 UID=1000
 GID=1001
--- a/datamodel/.docker/Dockerfile
+++ b/datamodel/.docker/Dockerfile
@@ -2,9 +2,9 @@
 
 # arm builds are not available with 3.2
 #FROM imresamu/postgis-arm64:14-3.2 AS base-arm64
-FROM postgis/postgis:14-3.2 AS base-arm64
+FROM postgis/postgis:14-3.4 AS base-arm64
 
-FROM postgis/postgis:14-3.2 AS base-amd64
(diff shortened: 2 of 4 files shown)


ponceta:
Can you propose a variable settings to define used postgres and postgis values? 

Because with the actual fix we would already have two versions used:

https://github.com/teksi/wastewater/blob/6cc3d8dfc220900cf18df602eabc2e9eefb4ace7/plugin/teksi_wastewater/interlis/utils/various.py#L87

If not required please just change the mentionned version to the same version as proposed.

ponceta (changes requested):
https://github.com/teksi/wastewater/blob/6cc3d8dfc220900cf18df602eabc2e9eefb4ace7/plugin/teksi_wastewater/interlis/utils/various.py#L87

sjib:
https://github.com/QGEP/QGEP/issues/821#issuecomment-2019993175

3nids:
@domi4484 do you have an idea how to synchronize the docker image in the plugin?

3nids:
good move @domi4484 
@ponceta if it's ok for you, you can approve and merge

ponceta (approved):
Thank you for the proposed corrections

Issue #192, opened 2024-04-08
Upgrade psycopg2 to psycopg3


ponceta:
As @3nids propose, we should try the upgrade to psychopg3 to ease the coding of complex views. (#149 )

https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#multiple-statements-in-the-same-query

https://www.psycopg.org/psycopg3/

3nids:
done in #194

PR #193, opened 2024-04-08
Bruteforce propose psycopg3

ponceta:
Bruteforce propose psycopg3

_Fun fact: there is no psycopg3 package!_

Fix https://github.com/teksi/wastewater/issues/192

Requires :

- [ ] Pirogue upgrade to psychopg3


ponceta:
SRID is not passing as it was.
`2024-04-08 13:49:06.763 UTC [176] ERROR:  type modifiers must be simple constants or identifiers at character 2170`
`, ST_Force2D(COALESCE(wn.situation3d_geometry, main_co.situation3d_geometry))::geometry(Point, $1) AS situation3d_geometry`

3nids:
superseded by #194

PR #196, opened 2024-04-09
Fix QGIS project details

ponceta:
- [x] Add polymorphic relations
 - [x] Fix reach cut
 - [x] Upgrade to QGIS 3.34.4
 - [ ] Fix digitize action


ponceta:
@urskaufmann @sjib The default identifier on wastewater nodes and covers is the obj_id of the wastewater structure, is it meant that way?

ponceta:
- Ok, yes it is the identifier of the wastewater structure which is equal to the obj_id IF the user didn't specify a name when creating the wastewater structure.
- And if I add an additionnal cover or node it will get the identifier specified by the user or the default one which is the obj_id of the additionnal cover or node.

**Small warning** : on actual state, if I rename my wastewater strucure identifier, this won't be inherited by the main cover and main node. Do we need to propose something to correct this?

image
Fig.1 : main cover and main node

image
Fig.2 : additionnal cover

urskaufmann:
Good hint, its one of my wishes:
If the user changes the identifier of the wastewater structure and the identifier of the node was before the same as the identifier of the structure, then the identifier of the node should als change. The same with the cover. Means: if the identifiers are same before, they should also be the same after changing the structure identifier.

ponceta:
image

ponceta:
ON QGEP we actually have
image

ponceta:
> ON QGEP we actually have image

I get the very same error as TWW if I explicitly set topology tool ON and allow intersects to happen.

ponceta:
Digitize : Add part on CurvePolygonZ is broken on QGIS:
https://github.com/qgis/QGIS/issues/57255

cymed:
Currently lacking the rights to update QGIS to 3.34.4 for testing

ponceta:
@cymed /!\ Digitize will not work until QGIS 3.34.6 is released, if possible, delay this upgrade at least until end of May 2024 for production topics. @cymed /!\ Digitize will not work until QGIS 3.34.6 is released, if possible, delay this upgrade at least until end of May 2024 for production topics.

sjib:
@ponceta One more thing to check and adapt: 
- [ ] And I wonder, why there is a link in the project - layer-properties vw_tww_damage for damage.single_damage_class. Imo a value relation is enough
see https://github.com/teksi/wastewater/issues/224

sjib:
To adapt also in project file `manhole_channel_distance` to `manhole_distance` (see https://github.com/teksi/wastewater/pull/230)

`	Zeile  73028:         <field configurationFlags="None" name="manhole_channel_distance">
	Zeile  73155:         <alias field="manhole_channel_distance" index="18" name=""></alias>
	Zeile  73186:         <default applyOnUpdate="0" expression="" field="manhole_channel_distance"></default>
	Zeile  73217:         <constraint constraints="0" exp_strength="0" field="manhole_channel_distance" notnull_strength="0" unique_strength="0"></constraint>
	Zeile  73248:         <constraint desc="" exp="" field="manhole_channel_distance"></constraint>
	Zeile  73286:           <column hidden="0" name="manhole_channel_distance" type="field" width="-1"></column>`

sjib:
- [ ] one more thing to fix ` 2infra/infraprogs/QGIS 3.0/bin:` see https://github.com/teksi/wastewater/issues/226#issuecomment-2083295918

PR #198, opened 2024-04-09
add psycopg2 support

3nids:
I just realized that we probably don't have support for psycopg3 under certain environment for now (OSGeo4W at least).

So, we have to make the plugin compatible with psycopg2 at least.
We will leave the test to work on psycopg2 as it's the most critical environment.


ponceta:
What tree do we have to shake to include psycopg (psycopg3) to future OSGeo4W versions?

ponceta:
https://trac.osgeo.org/osgeo4w/ticket/825#ticket opened an enhancement proposal here.

ponceta (approved):
👍

PR #203, opened 2024-04-12
Mask password in logs when calling interlis utilities

### LINKED ISSUE #167: Interlis export log include credentials in cleartext
When exporting a model from the GUI, the logs include credentials in cleartext on ili2pg calls.

domi4484:
Fix #167

### DIFF
--- a/plugin/teksi_wastewater/interlis/utils/various.py
+++ b/plugin/teksi_wastewater/interlis/utils/various.py
@@ -53,7 +53,7 @@ class CmdException(BaseException):
 
 
 def exec_(command, check=True, output_content=False):
-    command_masked_pwd = re.sub(r"(--dbpwd)\s\"[\w\.*#?!@$%^&-]+\"", r'\1 "[PASSWORD]"', command)
+    command_masked_pwd = re.sub(r"(--dbpwd)\s\'.+\'", r"\1 '[PASSWORD]'", command)
     logger.info(f"EXECUTING: {command_masked_pwd}")
     try:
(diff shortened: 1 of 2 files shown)


ponceta:
Add a comment?
```
# This command replaces the password in the command string with "[PASSWORD]" 
# to avoid exposing sensitive information in the logs, then saves the result with the level of information.
```

ponceta (approved):

plugin/teksi_wastewater/interlis/utils/various.py
```diff
@@ -52,7 +53,8 @@ class CmdException(BaseException):
 
 
 def exec_(command, check=True, output_content=False):
-    logger.info(f"EXECUTING: {command}")
+    command_masked_pwd = re.sub(r"(--dbpwd)\s\"[\w\.*#?!@$%^&-]+\"", r'\1 "[PASSWORD]"', command)
```
3nids, line 56: What about this to be safe?
```suggestion
    command_masked_pwd = re.sub(r"(--dbpwd)\s\".+\"", r'\1 "[PASSWORD]"', command)
```

3nids (approved):

plugin/teksi_wastewater/interlis/utils/various.py
3nids: and is the password is always in double quotes?
i.e. is there a risk of $ sign substitution if it's in double rather than single quotes? (which would an issue on its own)

PR #207, opened 2024-04-16
include oid_prefix in interlis text class export

cymed:
This PR 

- renames model_tww_vl to model_tww in the model mapping 
- adds a class for tww_sys in model_tww
- includes a query to tww_sys in the interlis export of labels

### DIFF
--- a/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py
+++ b/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py
@@ -2402,7 +2402,7 @@ class InterlisExporterToIntermediateSchema:
         return val
 
     def get_oid_prefix(self, oid_table):
-        instance = self.tww_session.query(oid_table).filter(oid_table.active == True).first()
+        instance = self.tww_session.query(oid_table).filter(oid_table.active is True).first()
         if instance is None:
             logger.warning(
                 f'Could not find an active entry in table"{oid_table.__table__.schema}.{oid_table.__name__}". Setting to default instead.'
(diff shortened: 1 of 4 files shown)


domi4484 (approved):
Looks good to me!

plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py
```diff
@@ -2405,9 +2405,12 @@ def get_oid_prefix(self, oid_table):
         instance = self.tww_session.query(oid_table).filter(oid_table.active is True).first()
         if instance is None:
             logger.warning(
-                f'Could not find an active entry in table"{oid_table.__table__.schema}.{oid_table.__name__}". Setting to default instead.'
+                f'Could not find an active entry in table"{oid_table.__table__.schema}.{oid_table.__name__}". \
+                Returning an empty string, which will lead to Interlis Errors. \
+                Set the value \'active\' in table"{oid_table.__table__.schema}.{oid_table.__name__}" to \
```
sjib: Set one value active that you want to use as prefix.

What happens if there is ore than one active?

cymed: with more than one entry set to active, the export takes the first one. This is almost never the case because more than one active entry corrupts oid generation. Fix will be needed once we support multitenancy

cymed:
@ponceta ready for merge

plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_exporter_to_intermediate_schema.py
sjib: I don't want to support multiple active prefixes at the moment, but to make sure that there are not several set to active - I open a new issue for that https://github.com/teksi/wastewater/issues/210

PR #211, opened 2024-04-17
Fix Re-Enable of triggers too early

domi4484:


ponceta (approved):
Successful import from @sjib, looks good to me!

Issue #117, opened 2024-01-29
Sequence number not correct after import


urskaufmann:
As we have already found a solution for the sequence number in qgep (https://github.com/QGEP/QGEP/issues/788) there is the same problem after migrating a qgep_database to tww_database: the sequence numbers must be updated (or we change the oid_prefix).
I tried to adapt the function and sql from QGEP issue 788, but get an error because of tww_od.network_node does not exist (materialized views are in tww_app instead of tww_od).
Should be fixed and the function installed in the database before users start migrating ...

cymed:
The problem is that that function can only be called **after** an import from dss. My suggestion is to migrate the data into the same database as before (as there is no overlap in schema names that is no problem) and then call

```
SELECT SELECT setval(tww.schemaname'.'tww.sequencename, qgep.last_value, true)
FROM
(SELECT * --sequence_schema, sequence_name 
FROM pg_sequences
WHERE schemaname='tww_od'
) tww
LEFT JOIN
(SELECT * --sequence_schema, sequence_name 
FROM pg_sequences 
WHERE schemaname='qgep_od'
)qgep on qgep.sequencename=tww.sequencename
WHERE qgep.last_value is not null;
```

urskaufmann:
we will use this function also after import of not teksi-data.

The problem is not that tww_od.network_node not exists, but tww_od.network_node has no obj_id-field (just an id-field). In QGEP, the table was in qgep_network - schema and was therefor not a problem. 
It works like this:

do
$$
declare
    f record;
	mytables_rec record;
begin
    for f in SELECT tablename, hastriggers 
       FROM pg_catalog.pg_tables
       WHERE schemaname = 'tww_od' and NOT tablename like 'network_%' and NOT tablename like 're_%'
	   ORDER BY tablename
    loop 
	raise notice '%', f.tablename;
	-- SELECT tww_sys.adjust_max_sequence ('mutation');
	EXECUTE 'SELECT tww_sys.adjust_max_sequence '
	  || '('''
      || f.tablename
	  || ''')'
	  INTO mytables_rec;
    end loop;
end;
$$

My solution: add the function to the datamodell and let the user start the adjust.sql (automatic after import and if oid_prefix is changed)
[adjust_max_sequence_tww.zip](https://github.com/teksi/wastewater/files/14118207/adjust_max_sequence_tww.zip)


By the way: the sequence-nr of this network-tables can/should be reset to 1 everytime the network topology is refreshed

[...]

PR #223, opened 2024-04-25
set obj_id sequences when importing

cymed:
This PR sets the sequences used for obj_id generation based on the current value of the sequence and the highest imported postfix with a matching shortcut. Mainly necessary when importing a dataset that was exported by TEKSI. Tackles #117


sjib:
@cymed Funktioniert das auch wenn du Daten mit verschiedenen Präfixes hast in einer Tabelle?

urskaufmann:
I do not really understand the code. I think to understand, that the prefix has no effect !?
I think that is not efficient. The postfix has to be unique per prefix and not overall.
If I choose a new prefix (what reason ever - because there are to many records), I should have a function who resets the sequence numbers. If I have to migrate my old database (that has three prefixes, two from the merged communities and one for the cantons road_drainage), I want the sequence_number for new records to be dependent on the actual prefix.

cymed:
> I do not really understand the code. I think to understand, that the prefix has no effect !? I think that is not efficient. The postfix has to be unique per prefix and not overall. If I choose a new prefix (what reason ever - because there are to many records), I should have a function who resets the sequence numbers. If I have to migrate my old database (that has three prefixes, two from the merged communities and one for the cantons road_drainage), I want the sequence_number for new records to be dependent on the actual prefix.

The latest push includes the oid_prefix and re-sets all sequences to the new prefix once there is change in tww_sys.oid_prefixes, making it more dynamic.

ponceta:
Same question as @sjib, what happens when you import data with a different oid than the active one?

sjib:
- [ ] I think it is right that the obj_id sequence is reset with taking into account the active prefix - because new data will be created using that prefix.

- [ ] If you change the prefix - then the obj_Id sequence again is adapted - and you can continue to create data with that new prefix.

So whether we import data with different prefixes or not seems not relevant here. More relevant is that it is clear with which prefix the user wants to go on creating new data.

- [ ] Updating obj_id sequence after each import I think is a good thing to implement - then we are sure that if any new data with the chosen prefix is added the obj_id sequence is adjusted before going on with adding data (original need from Urs in #117)

- [ ] So maybe we should include this fonctionality also on to a a new button or have some visualisation what prefix is currently acitve - see https://github.com/teksi/wastewater/pull/319

@cymed @urskaufmann @ponceta - are we here all at the same point of understanding?

cymed:
> * [ ]  I think it is right that the obj_id sequence is reset with taking into account the active prefix - because new data will be created using that prefix.
> 
>     * [ ]  If you change the prefix - then the obj_Id sequence again is adapted - and you can continue to create data with that new prefix.
> 
> 
> So whether we import data with different prefixes or not seems not relevant here. More relevant is that it is clear with which prefix the user wants to go on creating new data.
> 
>     * [ ]  Updating obj_id sequence after each import I think is a good thing to implement - then we are sure that if any new data with the chosen prefix is added the obj_id sequence is adjusted before going on with adding data (original need from Urs in [Sequence number not correct after import #117](https://github.com/teksi/wastewater/issues/117))
> 
>     * [ ]  So maybe we should include this fonctionality also on to a a new button or have some visualisation what prefix is currently acitve - see [Ensure that triggers stay active in case of error #319](https://github.com/teksi/wastewater/pull/319)
> 
> 
> @cymed @urskaufmann @ponceta - are we here all at the same point of understanding?

I agree in general. If we want to show which prefix is active, Also, once we go towards multitenancy, we'll need a way to switch between tenants anyway

urskaufmann:
I agree with @sjib . And to show the active prefix is also good idea (anyway,  if you care, you will see the active prefix with every new record you create in the obj_id...)

cymed:
@cymed dependecies on merging #381

sjib (commented):
Would it be wise to add some tests for that also?

cymed:
> Would it be wise to add some tests for that also?

What could break it? I don't have the time for creating the tests...

ponceta (approved):

PR #231, opened 2024-05-06
[WIP] Add urgency score and condition score weights

cymed:
This PR adds a sql script to embed the urgency score (and eventually the condition score) weights to the corresponding value lists

Progress

- [X] Add urgency score weights for channel attributes
- [ ] Define how to proceed with aquifier attributes
- [ ]  Add urgency score weights for aquifier attributes
- [ ] Add condition score weights for KEK attributes
- [ ] Implement solution on the plugin side to calculate and urgency scores from KEK import

### DIFF
--- a/datamodel/changelogs/0001/12_condition_score_urgency_parameters.sql
+++ b/datamodel/changelogs/0001/12_condition_score_urgency_parameters.sql
@@ -1,0 +1,48 @@
+/*
+Adding the urgency parameters and the structure condition parameters to the database
+*/
+-- this column is an extension to the VSA data model and assigns ugency weight to function hierarchic
+ALTER TABLE tww_vl.channel_function_hierarchic ADD COLUMN tww_weight_urgency smallint;
+UPDATE tww_vl.channel_function_hierarchic
+SET tww_weight_urgency= dict->>code
+FROM (
+	SELECT
+	jsonb_build_object(
+    '5068', 1,             -- pwwf.water_bodies
+    '5070', 0.9,           -- pwwf.main_drain_regional
+    '5069', 0.95,          -- pwwf.main_drain
+    '5071', 1,             -- pwwf.collector_sewer
+    '5062', 1,             -- pwwf.renovation_conduction
+    '5064', 1.1,           -- pwwf.residential_drainage
+    '5072', 1,             -- pwwf.road_drainage
+    '5066', 1,             -- pwwf.other
+    '5074', 1,             -- pwwf.unknown
+    '5063', 1,             -- swwf.renovation_conduction
+    '5065', 1,             -- swwf.residential_drainage
+    '5073', 1,             -- swwf.road_drainage
+    '5067', 1,             -- swwf.other
+    '5075', 1              -- swwf.unknown
+) dict )as obj
+;
+COMMENT ON COLUMN tww_od.channel_function_hierarchic.tww_weight_urgency IS 'this column is an extension to the VSA data model and assigns urgency weight
+ according to the VSA guidelines';
+ 
+ -- this column is an extension to the VSA data model and assigns ugency weight to function hierarchic
+ALTER TABLE tww_vl.channel_usage_current ADD COLUMN tww_weight_urgency smallint;
+UPDATE tww_vl.channel_usage_current
+SET tww_weight_urgency= dict->>code
+FROM (
+	SELECT jsonb_build_object(
+               '4526', .95,           -- wastewater
+               '4522', 1,             -- combined_wastewater
+               '4516', 1,             -- discharged_combined_wastewater
+               '4524', 0.9,           -- industrial_wastewater
+               '4514', 1.1,           -- clean_wastewater
+               '9023', 1.05,          -- surface_water
+               '4518', 1.1,           -- creek_water
+               '4571', 0.95,          -- other
+               '5322', 1              -- unknown
+           ) dict )as obj
+;
+COMMENT ON COLUMN tww_od.channel_usage_current.tww_weight_urgency IS 'this column is an extension to the VSA data model and assigns urgency weight
+ according to the VSA guidelines';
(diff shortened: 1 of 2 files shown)


cymed:
Suggestion to technical group: 

- Add a postgres function to calculate condition scores from latest KEK observation 
- Add a processing tool to access external layers for groundwater levels and protection zones and calculate weights in python code for these two parameters

sjib:
Background information:
Richtlinie «Zustandsbeurteilung von Entwässerungsanlagen»

Während gut zwei Jahre überarbeitete eine Arbeitsgruppe die Richtlinie «Zustandsbeurteilung von Entwässerungsanlagen» aus dem Jahr 2007. Neu wird jedem Schadencode des Merkblattes «Optische Inspektion von Entwässerungsanlagen: Schadencodierung und Datentransfer» eine Zustandsklasse zugewiesen. Zudem wird dargelegt, in welcher Form die Zustandsklasse des Einzelschadens auf ein Objekt (Kanal/Schacht) übertragen wird. Gleichzeitig wurden das Vorgehen und die Berechnung der Zustandsbewertung auf die neusten europäischen Normen abgestimmt. 
 
Die neue Richtlinie ist im VSA Webshop auf Deutsch, Französisch und bald auch in Italienisch erhältlich.  
https://www.aquaetgas.ch/de/vsa-news/centres-de-comp%C3%A9tences/neue-richtlinien-kurse-und-vorbereitung-zum-kanalisationsforum-2025/

urskaufmann:
What I miss: there are just the damage_codes in the database (BBCA), and this codes are the same in every language, but are repeated for every language. But the text to the codes (what means BBCA?) is nowhere. Would it be possible, the have the code just in the value_en - field, and in the value_de and value_fr fields the text for this code. It is then an exception, but it would help the user a lot.

urskaufmann:
To calculated the condition_score (Zustandsnote, vorher Baulicher Zustand) of every damage is the first thing to do. Following the new VSA Richtlinie, the lowest condition_score defines the class of the condition_score and the urgency_score (Dringlichkeitszahl), if you are just interested in the 5 mainclasses (dringend, kurzfristig, mittelfristig, langfristig, keine). 
Do we really need to calculate the condition_score or is it not sufficient to import this values from the tv_inspection_data? 
Aquifer_data: for me no priority. These data are to often missing.

cymed:
> What I miss: there are just the damage_codes in the database (BBCA), and this codes are the same in every language, but are repeated for every language. But the text to the codes (what means BBCA?) is nowhere. Would it be possible, the have the code just in the value_en - field, and in the value_de and value_fr fields the text for this code. It is then an exception, but it would help the user a lot.

Fiddling with value_de messes up INTERLIS import and export. We could add a field "description_en"/"description_de" in the database and use it in the value relations. Not part of the urgency score calculations though

cymed:
> To calculated the condition_score (Zustandsnote, vorher Baulicher Zustand) of every damage is the first thing to do. Following the new VSA Richtlinie, the lowest condition_score defines the class of the condition_score and the urgency_score (Dringlichkeitszahl), if you are just interested in the 5 mainclasses (dringend, kurzfristig, mittelfristig, langfristig, keine). Do we really need to calculate the condition_score or is it not sufficient to import this values from the tv_inspection_data? Aquifer_data: for me no priority. These data are to often missing.

Zustandsnote is not part of VSA-KEK, so I think we could add a column tww_damage_score to the damage value lists

sjib:
> What I miss: there are just the damage_codes in the database (BBCA), and this codes are the same in every language, but are repeated for every language. But the text to the codes (what means BBCA?) is nowhere. Would it be possible, the have the code just in the value_en - field, and in the value_de and value_fr fields the text for this code. It is then an exception, but it would help the user a lot.

For me this raises a general question if we should also add description fields to the values:

 INSERT INTO tww_vl.damage_channel_channel_damage_code (code, vsacode, value_en, value_de, value_fr, value_it, value_ro, abbr_en, abbr_de, abbr_fr, abbr_it, abbr_ro, active) VALUES (4046,4046,'BCAAA','BCAAA','BCAAA', 'BCAAA', 'BCAAA', '', '', '', '', '', 'true');

The descriptions exist in the INTERLIS and could be easily integrated also in TEKSI:

```
      !!@ comment = "Tier in der offenen Rohrverbindung"
      BBHZC,
      !!@ comment = "Tier"
      BBHZZ,
      !!@ comment = "Anschluss mit Formstück"
      BCAAA,
```

We could check if we then could display them:

- Within the tool tip
- on mouseover somewhere in the GUI
- other option for this in QGIS

@urskaufmann @cymed @ponceta and others - What do you think?

urskaufmann:
> Zustandsnote is not part of VSA-KEK, so I think we could add a column tww_damage_score to the damage value lists

This will not work, because the condition_score of a single damage is dependent also from the values in the channel_quantification1 and 2 fields. To store condition_score of a single damage, we have to add a new _field in table damage_channel, for condition_score of a channel or manhole, we have the attribut wastewater_structure.condition_score.

urskaufmann:
> We could check if we then could display them:
> 
> * Within the tool tip
> * on mouseover somewhere in the GUI
> * other option for this in QGIS

It would be also easy possible to add this text in a view or the copy the text in a maintenace event... ?

cymed:
> > Zustandsnote is not part of VSA-KEK, so I think we could add a column tww_damage_score to the damage value lists
> 
> This will not work, because the condition_score of a single damage is dependent also from the values in the channel_quantification1 and 2 fields. To store condition_score of a single damage, we have to add a new _field in table damage_channel, for condition_score of a channel or manhole, we have the attribut wastewater_structure.condition_score.

This in turn means that we'd need a good chunk of new functions on the database or in the plugin ( I think database is smarter)

urskaufmann:
> This in turn means that we'd need a good chunk of new functions on the database or in the plugin ( I think database is smarter)

--> is it really necessary to calculate this value in TWW, if we normally get the value form pallon or wincan or other protocolling tools?

sjib:
to come back to the original questions:
The canton Lucerne published there groundwater levels:
https://daten.geo.lu.ch/produkt/grwasxxx_col_v4
Are such datasets enough for the decision whether a pipe is within the groundwater level or not?

cymed:
not following up on the draft - reopen if needed

Issue #234, opened 2024-05-07
Translations are broken


ponceta:
Translations are broken

https://teksi.github.io/wastewater/fr/features-guide/index.html

Should be translated

ponceta:
https://github.com/teksi/wastewater/actions/runs/8980790102/artifacts/1479120618

ponceta:
Fix by https://github.com/teksi/wastewater/commit/1c4ca04dc0ee2b755ef8750d87ee2bd78d6ec73e

PR #235, opened 2024-05-07
Mini update to check documentation translation workflow

ponceta:
Mini update to check documentation translation workflow


ponceta:
https://github.com/teksi/wastewater/issues/234

Issue #232, opened 2024-05-06
Adapt organisation_functions.sql class name _aquifer to dss15_aquifer


sjib:
This reference on FROM tww_od._aquifier seems to be outdated:

`FROM tww_od._aquifier`
https://github.com/teksi/wastewater/blob/main/datamodel/app/organisation_functions.sql#L9

Table _aquifier was renamed to tww_od.dss15_aquifer
https://github.com/teksi/wastewater/blob/6c37125b886726697bea9c5997f3633761c9f059/datamodel/changelogs/0001/51_dss15_aquifer.sql#L8

Should be 
`FROM tww_od.dss15_aquifer`

sjib:
@cymed Can you confirm? Then I will make a pull request to change that.

cymed:
I can confirm

PR #238, opened 2024-05-07
Fix organisation_functions.sql

ponceta:
Fix https://github.com/teksi/wastewater/issues/232


sjib (approved):

sjib:
@ponceta Go ahead with this!

PR #239, opened 2024-05-07
Correct classname _aquifer to dss15_aquifer

sjib:
Should fix https://github.com/teksi/wastewater/issues/232


sjib:
Duplicate of https://github.com/teksi/wastewater/pull/238

PR #248, opened 2024-05-27
Setup databases extensions in first changelog

### LINKED ISSUE #247: Creation of extensions missing
In qgep we created the extensions, if they were missing in [00_qgep_schema.sql](https://github.com/QGEP/datamodel/blob/master/00_qgep_schema.sql)

```
CREATE EXTENSION IF NOT EXISTS postgis;
CREATE EXTENSION IF NOT EXISTS hstore;
```
I cannot find this anymore in the same script of TWW https://github.com/teksi/wastewater/blob/main/datamodel/changelogs/0001/01_schema.sql

@ponceta Any reason for that?
Where do we integrate these extensions in the database dump?

ponceta:
- [x] Setup databases extensions in first changelog

Should fix #247


3nids:
I'm not sure if that should be part of the changelogs, that's more part of the database setup and the requirements. But looking at the dumps, the cenxtensions creations are part of it, so...it sounds ok.

3nids (approved):

sjib:
@ponceta I usually also created `hstore` - do we need that for anything already?
https://www.postgresql.org/docs/current/hstore.html

cymed:
> @ponceta I usually also created `hstore` - do we need that for anything already? https://www.postgresql.org/docs/current/hstore.html

It is needed for the History viewer

datamodel/changelogs/0001/00_db_extensions.sql
```diff
@@ -1,2 +1,3 @@
 CREATE EXTENSION IF NOT EXISTS postgis;
+CREATE EXTENSION IF NOT EXISTS hstore;
```
3nids, line 2: The audit is not currently bundled. I would recommend to revisit this before adding this extension. Jon(b) is generally preferred over hstore nowadays.

cymed: dropping hstore requires changes in the history viewer plugin

3nids: Yes, but I am just saying that currently there is nothing in this repo related to the history viewer. It's up to the manager to add this extension, it should not be integrated by default if not use by default.

ponceta: https://github.com/teksi/wastewater/pull/249 I changed the extensions here and added the history viewer needs in a separate PR.

We can always switch to jsonb when we implement it in history viewer.

This is a 1-1 functionnality with QGEP.

ponceta: See https://github.com/teksi/wastewater/pull/249 for later comments.

PR #268, opened 2024-06-25
Add Layers necessary for AG-xx to project file

cymed:
This PR adds the layers necessary for AG-xx Export to the qgis project - namely the three measure layers and building group. Pushing the project changes directly to tww so we don't run into problems further down the line.

image


sjib:
@cymed Do you plan to integrate everything in the same project file of tww or do you plan to have an AG64/96 version of the project?

cymed:
> @cymed Do you plan to integrate everything in the same project file of tww or do you plan to have an AG64/96 version of the project?

I do not want to have a separate QGIS file if possible (more stuff to manage). If you open a project linked to an agxx-enabled db, all the additional fields are automatically loaded. The AG-xx users can then add the AG-xx forms for the relevant layers themselves

ponceta (approved):
Nice improvement of the current tww project

PR #275, opened 2024-07-01
use 2d length for length_effective if rp_level is zero

cymed:
When importing data that has zero values for rp_level, the reach.length_effective should not be calculated in 3D.


sjib:
@cymed What do yo do when the two levels are NULL? Should this be handeled the same way as if they are 0 also?

cymed:
> @cymed What do yo do when the two levels are NULL? Should this be handeled the same way as if they are 0 also?

This case was already handled by the coalsece in the else clause

sjib (approved):
Thanks a lot!

PR #217, opened 2024-04-23
2024-04-interlis-export-orientation-option

sjib:

PR #281, opened 2024-07-04
fix label export orientation

cymed:
Up to now, the labels_orientation_offset is not passed to the function _modulo_angle in the intermediate schema.

This PR takes the changes from #217 and applies the following fixes:

- tww:extractlabels_interlis has no Parameter "EXPORT_ORIENTATION". Parameter is dropped on export of labels. file.
- On Export, the orientation offset is passed to the intermediate schema
- Without this PR, the export of labels will not work


domi4484 (approved):

sjib:
@cymed Herzlichen Dank!

sjib (commented):
ok, thanks

PR #288, opened 2024-07-05
bump pirogue

### LINKED ISSUE #65: SingleInheritance views use default obj_id of parent instead of child
If the Primary Key of a SingleInheritance view (i.e. tww_app.vw_cover) is missing, the insert command from pirogue uses the PK of the parent (i.e. structure_part). It would be more user friendly to use the child's PK.

3nids:
fixes #65

image


cymed (approved):
Looks good

Issue #250, opened 2024-05-27
No Main-Node


urskaufmann:
I just tried what happens, if
- I delete the main_cover of a wastewater_structure with two covers: the remaining cover get's the new main_cover. Good!
- I delete the main_node of a wastewater_structure with two nodes:  the symbol goes to the existing cover (not node!), the main_node stays empty, I have no ui to define a new main_node. I can create a next node in tab wastewater_nodes, but I can not save, because the main_node in tab main_cover/main_node is empty.

There should be a ui to choose/define a main_node and when deleting the main_node, the user should get at least a warning.

ponceta:
Same behaviour for QGEP

ponceta:
A graphical way to do it in QGIS (in wastewater_structures)
1. Define a relation :
![image](https://github.com/teksi/wastewater/assets/8614223/043def73-1aba-40cc-bd0f-36644fa576c3)
2. Set relation reference widget for fk_main_wastewater_node
![image](https://github.com/teksi/wastewater/assets/8614223/606f38bd-19f6-4cea-9136-8c082127d9df)
3. Open the wastewater form and click on the map to define the main node
![image](https://github.com/teksi/wastewater/assets/8614223/e27ebedf-5c24-4649-8308-1cb872072894)

**Extra :**
If we want the selection to be limited to the wastewater nodes referencing the wastewater structure, we could set a filter to propose only these entities but I would test that in the expression workshop of next QGIS user day (I do not know the syntax yet).

**The simpler the better :**
If the logic exists for covers, make the definition of the fk_main_ws_node simply the next referencing ws_node.

sjib:
> If we want the selection to be limited to the wastewater nodes referencing the wastewater structure
I think this is a must else we have a chaos

sjib:
To think about (I have no preference yet):
my question is also if a graphical selection is helpful or if a list to choose from (that offers only the connected nodes resp. covers) would'nt be the easier interface?

ponceta:
> To think about (I have no preference yet): my question is also if a graphical selection is helpful or if a list to choose from (that offers only the connected nodes resp. covers) would'nt be the easier interface?

Even with a tabular selection, the user is still able to see where the entity lies on the map through redlining.

cymed:
We use sql functions on the db side after interlis imports, see #97. They are however not called by triggers. Also I am not fully sure about the behaviour for structures with more than 2 nodes.

ponceta:
[...]

PR #292, opened 2024-07-08
Add same behaviour as for cover on the fk_main cover

ponceta:
- [x] Add same behaviour as for cover on the fk_main cover
- [x] Reword syntax for node to avoid misunderstandings
- [  ] Require testing
- [  ] Review ready

Linked Issues

- [ ] Fix https://github.com/teksi/wastewater/issues/250

### DIFF
--- a/datamodel/app/symbology_functions.sql
+++ b/datamodel/app/symbology_functions.sql
@@ -641,24 +641,25 @@ CREATE OR REPLACE FUNCTION tww_app.on_wasterwaternode_change()
   RETURNS trigger AS
 $BODY$
 DECLARE
-  co_obj_id TEXT;
-  affected_sp RECORD;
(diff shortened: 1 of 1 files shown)


ponceta:
A test should be written :

1. Where a wastewater structure has 2 covers (1 main and 1 additionnal) and 2 nodes (1 main and 1 additionnal)
2. Remove the main cover
3. Assert that the wastewater structure defined the second cover as main.
4. Remove the main node
5. Assert that the wastewater structure defined the second node as main.

To avoid any regression in this behaviour.

ponceta:
This does not fix the issue yet.

`SELECT tww_app.wastewater_structure_update_fk_main_wastewater_node(NULL, true)`

Works on demo dataset

PR #300, opened 2024-07-10
Allow to load official extensions via shell

cymed:
As suggested in https://github.com/orgs/teksi/discussions/108, we need a way to include officially supported extensions in CI and deployment of TEKSI. As a first step, this PR creates an environment from which we can install extensions. Alterations to the CI itself are not done yet.

The solution works by calling `python %tww_dir%\extensions\extension_manager.py --pg_service %myservice% --module_name tww --srid 2056 --drop-schema --extension_name demo` from the shell. 
In a yaml, The key informations for creation (id, directory name, schema name and variables) for the extension are defined. The code then sorts all the files in the directory defined in the yaml and executes all .sql and .py scripts in that folder

Furthermore, there are now extension flags in create_dumps.py that alter the output file name (not productive yet as we have no CD invoking all processes for extensions as well)


cymed:
I fail to dynamically populate ${EXT_NAME} in setup.sh. Help wanted

cymed:
closing here, will be re-pushed via TEKSI 2 AG-64/96

PR #301, opened 2024-07-10
set organisation.tww_active after import

cymed:
After importing, we should limit tww_active to the values effectively used


sjib:
@cymed That is a nice new fonctionality. 
How can the user add further organisations as active that are not yet in the current data?
Do we there still go thru pg admin? 
Or do we have a layer and a form to adjust? 
What happens if I import more data - will it be re-set to inactive then or are already set to active organisations kept active?

Some documenation on this with this pr would be helpful.

cymed:
> @cymed That is a nice new fonctionality. How can the user add further organisations as active that are not yet in the current data? Do we there still go thru pg admin? Or do we have a layer and a form to adjust? What happens if I import more data - will it be re-set to inactive then or are already set to active organisations kept active?
> 
> Some documenation on this with this pr would be helpful.

They are still loaded into tww_od.organisation, so it can be altered in the corresponding QGIS layer

cymed:
I added a flag in the sql script. Now, the existing active values are kept

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -140,6 +140,9 @@ def interlis_import(self, xtf_file_input, show_selection_dialog=False, logs_next
         self._progress_done(95, "Update main cover and refresh materialized views...")
         self._import_update_main_cover_and_refresh_mat_views()
 
+        # Update main_cover and main_wastewater_node
```
ponceta, line 143: Wrong comment

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -256,6 +259,18 @@ def _import_from_intermediate_schema(self, import_model):
 
         return interlisImporterToIntermediateSchema.session_tww
 
+    def _import_manage_organisations(self):
```
ponceta, line 262: Is manage a good term for this function?

ponceta:
If I understand correctly you loop on every od table and set every organisation tww_active to TRUE if you encounter these in the data.

- [ ] Just make sure you don't deactivate already active organisations when doing a small import to an existing database.

Is this fast on a large dataset?

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
ponceta, line 143:
```suggestion
        # Update organisations
```

ponceta (approved):

PR #304, opened 2024-07-11
2024-07-11-interlis-import-fix-references-value-lists

sjib:
- [x] measuring_point - purpose & damming_device - treat as value lists as in data model definition

### DIFF
--- a/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
+++ b/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
@@ -2063,9 +2063,9 @@ class InterlisImporterToIntermediateSchema:
                 data_details=row.detaildaten,
                 duration=row.dauer,
                 # in VSA-KEK 2020 in class maintenance_event instead of examination
-                #fk_operating_company=(
+                # fk_operating_company=(
                 #    row.ausfuehrende_firmaref if row.ausfuehrende_firmaref else None
-                #),
+                # ),
                 identifier=row.bezeichnung,
                 operator=row.ausfuehrender,
                 reason=row.grund,
--- a/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
+++ b/plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
@@ -684,7 +684,8 @@ class InterlisImporterToIntermediateSchema:
                     self.model_classes_tww_od.discharge_point_relevance, row.relevanz
                 ),
                 terrain_level=row.terrainkote,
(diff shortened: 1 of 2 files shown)


sjib:
to adapt:

- [ ]                 # fk_disposal=self.get_pk(row.entsorgungref__REL), # TODO check why not available
Has to be matched to re_table

sjib:
- [x] maintenance_event.active_zone - old extension of qgep - not supported anymore in tww

```
            logger.warning(
                "TWW examination.active_zone has no equivalent in the interlis model. This field will be null."
            )

                # --- maintenance_event ---
                # active_zone=row.REPLACE_ME,  # TODO : found no matching field for this in interlis, confirm this is ok

```

sjib:
Untersuchung - nicht mehr notwendig
      ```
          fk_operating_company=(
                    row.ausfuehrende_firmaref if row.ausfuehrende_firmaref else None
                ),
```

**Comment by sjib** [comment]
- [ ] to correct vsa-kek 2020 - attributes moved to super class / einige attribute wurden in oberklasse verschoben

    def _import_schaden und kanalschaden und normschachtschaden anpassen

**Comment by sjib** [comment]
@domi4484 Can you have a look and review on this and check also the newly created issues 306 and 307 and comment on them

**Comment by sjib** [comment]
Should solve https://github.com/teksi/wastewater/issues/303

File: plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py [hunk 1]
```diff
@@ -445,6 +445,10 @@ def wastewater_structure_common(self, row):
             ),
             "contract_section": row.baulos,
             "detail_geometry3d_geometry": ST_Force3D(row.detailgeometrie),
+            # TODO : NOT MAPPED VSA-DSS 3D
```
ponceta, line 448: @sjib could you raise an issue for that and mention the requirements to make it 1:1 with VSA DSS 2020 3D?

sjib: There exists INTERLIS VSA-DSS 2020 3D - but it is not used currently as we have a "problem" in INTERLIS that we can create e.g. detail_geometry3d, but will not have 3D information in most cases for a whole community right now. And if we do not have that, then the validation of INTERLIS VSA-DSS 2020 3D fails. But it is also a little bit a henn - egg problem. I will raise a discussion as proposed.

ponceta (approved):

plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
sjib: https://github.com/orgs/teksi/discussions/109

Issue #305, opened 2024-07-12
VSA-KEK 2020.1 import - additional file created on import


sjib:
**Describe the bug**
When importing the following dataset with VSA-KEK data I get a new dataset file that is not in the xtf file.
[minimal-dataset-transferdatensatz_VSA-KEK-manhole-damage.zip](https://github.com/user-attachments/files/16193957/minimal-dataset-transferdatensatz_VSA-KEK-manhole-damage.zip)

(adaption of [plugin/teksi_wastewater/tests/data/minimal-dataset-VSA-KEK-manhole-damage.xtf](https://raw.githubusercontent.com/teksi/wastewater/main/plugin/teksi_wastewater/tests/data/minimal-dataset-VSA-KEK-manhole-damage.xtf) for the VSA Transferdatensatz) with AbwasserbauwerkRef replaced `AbwasserbauwerkRef REF="ch000000WS000001"` with `AbwasserbauwerkRef REF="ch080qwzNS000067"` - this is Abwasserbauwerk with identifier `1.040`

**To Reproduce**
Exact steps to reproduce the behavior:
1. Use latest TEKSI Wastewater and import organisations and Transferdatensatz VSA
2. Go to 'INTERLIS Import'
3. Select new dataset `minimal-dataset-transferdatensatz_VSA-KEK-manhole-damage.xtf` to add VSA-KEK data to test VSA-KEK import
4. Launch import
5. Adapt absolute file path for data_media, to fit your local system and save with "Set"
6. 
![20240712_import_kek_adapt_path2_set_ok](https://github.com/user-attachments/assets/76aea7cf-e5e4-4d39-be62-b2aff9ce482c)
![20240712_import_kek_adapt_path](https://github.com/user-attachments/assets/5a7f46e9-4594-4314-950a-9118ecfd7301)

7. Finish import process
8. Check entries in table file

**Expected behavior**
_A clear and concise description of what you expected to happen._

No additional file dataset [ch000000FI000002] should be generated on import of VSA-KEK 2020 data

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.
![20240712_kek_import_additional_file_dataset_created](https://github.com/user-attachments/assets/9073a77a-ae1d-4fa2-96d1-e546bc600a00)

**Desktop (please complete the following information):**
 - TWW version
 - QGIS Version
 - OS

**Additional context**

sjib:
[...]

PR #308, opened 2024-07-12
Add VSA-KEK manhole damage dataset that fits VSA Transferdatensatz 

sjib:
instead of Arbon data) to be able to test KEK import there


sjib:
see also Test VSA-KEK 2020 INTERLIS Import https://github.com/teksi/wastewater/issues/305

ponceta (approved):

PR #322, opened 2024-07-18
QGIS project update

ponceta:
- [x] Upgrade project to QGIS 3.34.8
 - [x] Add tww_od.default_values to the project
 - [x] Add tww_app.vw_wastewater_structure instead of wastewater_structure


cymed:
Works as expected when using 3.34.8. Do we check the minimal QGIS version somewhere? 
I went over the attribute form and deleted some fields that are not there. here is a qml to alter it (had to save it as txt to upload it on github.
[vw_wastewater_structure.txt](https://github.com/user-attachments/files/16284872/vw_wastewater_structure.txt)

ponceta:
> Works as expected when using 3.34.8. Do we check the minimal QGIS version somewhere?

No but if you open it with an earlier version you will get a warning from QGIS.

cymed:
Altered forms ok for you?

cymed (approved):

PR #325, opened 2024-07-19
[WIP] Add listener to DatabaseUtils

cymed:
In order to push warnings etc. from the db to QGIS iface (as needed i.e. in #36), we need to have a plugin side option to listen to pg_notify events. I try to get there via this draft PR, yet it needs technical validation (to make sure we properly close the connections every time).

### planned usage

add pg_notify in a trigger function on the database side, i.e. the on insert of wastewater structure

https://github.com/teksi/wastewater/blob/a7dc88240bbfd8e45b68b5b305c1f7633a2bf90e/datamodel/app/view/vw_tww_wastewater_structure.py#L276-L284

On the plugin side, we connect the Channel 'vw_tww_ws_no_cover' using

```
no_cover_payload = DatabaseUtils.dblisten('vw_tww_ws_no_cover')
for payload in no_cover_payload:
    self.iface.messageBar().pushMessage(
                "Warning",
                payload,
                level=Qgis.Warning,
            )
```

### DIFF
--- a/plugin/teksi_wastewater/teksi_wastewater_plugin.py
+++ b/plugin/teksi_wastewater/teksi_wastewater_plugin.py
@@ -155,7 +155,7 @@ class TeksiWastewaterPlugin:
             self.iface.mainWindow(), ["vw_wastewater_node", "vw_tww_reach"]
         )
         self.wastewater_structure_layer_notifier = TwwLayerNotifier(
-            self.iface.mainWindow(), ["vw_wastewater_structure"]
+            self.iface.mainWindow(), ["vw_tww_wastewater_structure"]
         )
 
         self.toolbarButtons = []
--- a/plugin/teksi_wastewater/teksi_wastewater_plugin.py
+++ b/plugin/teksi_wastewater/teksi_wastewater_plugin.py
@@ -157,7 +157,7 @@ class TeksiWastewaterPlugin:
         self.wastewater_structure_layer_notifier = TwwLayerNotifier(
             self.iface.mainWindow(), ["vw_wastewater_structure"]
         )
-        
+
         self.toolbarButtons = []
 
         # Create toolbar button
@@ -338,7 +338,7 @@ class TeksiWastewaterPlugin:
         self.wastewater_networkelement_layer_notifier.layersAvailableChanged.connect(
             self._wastewater_networkelement_layer_available_changed
         )
-        
+
         self.wastewater_structure_layer_notifier.layersAvailableChanged.connect(
             self._wastewater_structure_layer_available_changed
         )
@@ -657,26 +657,26 @@ class TeksiWastewaterPlugin:
             self._configure_database_connection_config_from_tww_layer()
 
             self.tww_validity_check_startup()
-            no_cover_payload = DatabaseUtils.dblisten('vw_tww_ws_no_cover')
+            no_cover_payload = DatabaseUtils.dblisten("vw_tww_ws_no_cover")
             for payload in no_cover_payload:
                 self.iface.messageBar().pushMessage(
-                            "Warning",
-                            payload,
-                            level=Qgis.Warning,
-                        )
-                        
+                    "Warning",
+                    payload,
+                    level=Qgis.Warning,
+                )
+
     def _wastewater_structure_layer_available_changed(self, available):
 
         if available:
             self._configure_database_connection_config_from_tww_layer()
 
-            no_cover_payload = DatabaseUtils.dblisten('vw_tww_ws_no_cover')
+            no_cover_payload = DatabaseUtils.dblisten("vw_tww_ws_no_cover")
             for payload in no_cover_payload:
                 self.iface.messageBar().pushMessage(
-                            "Warning",
-                            payload,
-                            level=Qgis.Warning,
-                        )
+                    "Warning",
+                    payload,
+                    level=Qgis.Warning,
+                )
 
     def update_admin_mode(self):
 
--- a/plugin/teksi_wastewater/teksi_wastewater_plugin.py
+++ b/plugin/teksi_wastewater/teksi_wastewater_plugin.py
@@ -657,13 +657,6 @@ class TeksiWastewaterPlugin:
             self._configure_database_connection_config_from_tww_layer()
 
             self.tww_validity_check_startup()
-            no_cover_payload = DatabaseUtils.dblisten("vw_tww_ws_no_cover")
-            for payload in no_cover_payload:
-                self.iface.messageBar().pushMessage(
-                    "Warning",
-                    payload,
-                    level=Qgis.Warning,
-                )
 
     def _wastewater_structure_layer_available_changed(self, available):
 
--- a/plugin/teksi_wastewater/teksi_wastewater_plugin.py
+++ b/plugin/teksi_wastewater/teksi_wastewater_plugin.py
@@ -154,6 +154,10 @@ class TeksiWastewaterPlugin:
         self.wastewater_networkelement_layer_notifier = TwwLayerNotifier(
             self.iface.mainWindow(), ["vw_wastewater_node", "vw_tww_reach"]
         )
+        self.wastewater_structure_layer_notifier = TwwLayerNotifier(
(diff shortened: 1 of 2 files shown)


plugin/teksi_wastewater/teksi_wastewater_plugin.py
```diff
@@ -154,6 +154,10 @@ def initGui(self):
         self.wastewater_networkelement_layer_notifier = TwwLayerNotifier(
             self.iface.mainWindow(), ["vw_wastewater_node", "vw_tww_reach"]
         )
+        self.wastewater_structure_layer_notifier = TwwLayerNotifier(
+            self.iface.mainWindow(), ["vw_wastewater_structure"]
```
domi4484, line 158: This is not working (at least for me the slot never get called) because in the project there is no layer named `vw_wastewater_structure`
If you need only a project opened/closed slot you could use `_wastewater_networkelement_layer_available_changed` without making a new listener

plugin/teksi_wastewater/utils/database_utils.py
```diff
@@ -220,3 +221,22 @@ def get_validity_check_issues() -> List[str]:
...
+                cursor.execute(f"LISTEN {channel}")
+                try:
+                    while True:
+                        if select.select([connection], [], [], timeout) == ([], [], []):
+                            print(f"Timeout on Channel {channel}")
+                        else:
+                            connection.poll()
+                            while connection.notifies:
+                                notify = connection.notifies.pop()
+                                yield notify.payload
```
domi4484, line 239: This part is blocking the gui thread and makes qgis freeze. It should be moved into a separate thread and communicate with the gui via signal/slots

domi4484 (commented):
@cymed I tried that out but doesn't work for me. I see you added the label "help needed", do you need some guidance or should I take over this PR?

What operation can you do in the project to get the pg_notify raised?

plugin/teksi_wastewater/teksi_wastewater_plugin.py
cymed: I will alter it to vw_tww_wastewater_structure

cymed:
> @cymed I tried that out but doesn't work for me. I see you added the label "help wanted", do you need some guidance or should I take over this PR?

I have no experience regarding the QGIS side of the pg_notify world. Help concerning the pyqt integration is appreciated

> 
> What operation can you do in the project to get the pg_notify raised?

once #36 is merged, I pg_notify after adding an entry in tww_app.vw_tww_wastewater_structure with no cover attributes (except co_obj_id and co_identifier). In these cases we do not add a cover on the db and would like to inform the user about it. 

See 
https://github.com/teksi/wastewater/blob/a7dc88240bbfd8e45b68b5b305c1f7633a2bf90e/datamodel/app/view/vw_tww_wastewater_structure.py#L269-L284

domi4484:
One issue I see with this approach is that when someone add a feature with no cover, all other persons with the project open will receive the notification. But maybe it is not an issue.

I would probably implement something like this (disclaimer: not tested and I never used pg_notify before):
```python
class DatabaseNotifyListener(QThread):

    # Timeout every seconnd, this is the max time that the gui thread could be 
    # blocked when stopping the listener
    _SELECT_TIMEOUT = 1.0
    _STOP_TIMEOUT = 2.0

    # signal to communicate with main thread
    payloadSignal = pyqtSignal(str)

    def __init__(self, channel: str):
        super().__init__(None)
        self._channel = channel
        self._stop = False

    def startListener(self):
        # check already running
        if self.isRunning():
            self.stopListener()

        self._stop = False

        # Start the QThread, run() will be called whitin the thread
        self.start()

    def stopListener(self):
        self._stop = True

        # wait for the thread to be stopped
        if not self.wait(self._STOP_TIMEOUT):
            # Raise or print error that thread could not be stopped
            pass

    def run(self):
        with DatabaseUtils.PsycopgConnection() as connection:
            try:
                cursor = connection.cursor()
                cursor.execute(f"LISTEN {self._channel}")
                while True:
                    if self._stop:
                        break

                    if select.select([connection], [], [], self._SELECT_TIMEOUT) == ([], [], []):
                        continue

                    connection.poll()
                    while connection.notifies:
                        if self._stop:
                            break

                        notify = connection.notifies.pop()
                        self.payloadSignal.emit(notify.payload)

            finally:
                cursor.execute(f"UNLISTEN {self._channel}")
                cursor.close()
```

Usage from `teksi_wastewater_plugin.py`:

```python
def initGui():
    self.dbNotifyListenerNoCover = DatabaseNotifyListener("vw_tww_ws_no_cover")
    self.dbNotifyListenerNoCover.payloadSignal.connect(self.dbNotifyListenerNoCoverPayloadAvailable)

def unload():
    if self.dbNotifyListenerNoCover.isRunning():
        self.dbNotifyListenerNoCover.stopListener()

def _wastewater_structure_layer_available_changed(self, available):
    if available:
        self.dbNotifyListenerNoCover.startListener()
    else:
        self.dbNotifyListenerNoCover.stopListener()

def dbNotifyListenerNoCoverPayloadAvailable(self, payload: str):
    self.iface.messageBar().pushMessage(
        "Warning",
        payload,
        level=Qgis.Warning,
        )
```

cymed:
> One issue I see with this approach is that when someone add a feature with no cover, all other persons with the project open will receive the notification. But maybe it is not an issue.
> 

Good point. It would probably be a better approach to use QGIS actions then. I do see other cases where we can use this approach feasibly (i.e. on alterations of default value).

> I would probably implement something like this (disclaimer: not tested and I never used pg_notify before):
> ```python
> class DatabaseNotifyListener(QThread):
> 
>     # Timeout every seconnd, this is the max time that the gui thread could be 
>     # blocked when stopping the listener
>     _SELECT_TIMEOUT = 1.0
>     _STOP_TIMEOUT = 2.0
> 
>     # signal to communicate with main thread
>     payloadSignal = pyqtSignal(str)
> 
>     def __init__(self, channel: str):
>         super().__init__(None)
>         self._channel = channel
>         self._stop = False
> 
>     def startListener(self):
>         # check already running
>         if self.isRunning():
>             self.stopListener()
> 
>         self._stop = False
> 
>         # Start the QThread, run() will be called whitin the thread
>         self.start()
> 
>     def stopListener(self):
>         self._stop = True
> 
>         # wait for the thread to be stopped
>         if not self.wait(self._STOP_TIMEOUT):
>             # Raise or print error that thread could not be stopped
>             pass
> 
>     def run(self):
>         with DatabaseUtils.PsycopgConnection() as connection:
>             try:
>                 cursor = connection.cursor()
>                 cursor.execute(f"LISTEN {self._channel}")
>                 while True:
>                     if self._stop:
>                         break
> 
>                     if select.select([connection], [], [], self._SELECT_TIMEOUT) == ([], [], []):
>                         continue
> 
>                     connection.poll()
>                     while connection.notifies:
>                         if self._stop:
>                             break
> 
>                         notify = connection.notifies.pop()
>                         self.payloadSignal.emit(notify.payload)
> 
>             finally:
>                 cursor.execute(f"UNLISTEN {self._channel}")
>                 cursor.close()
> ```
> 
> Usage from `teksi_wastewater_plugin.py`:
> 
> ```python
> def initGui():
>     self.dbNotifyListenerNoCover = DatabaseNotifyListener("vw_tww_ws_no_cover")
>     self.dbNotifyListenerNoCover.payloadSignal.connect(self.dbNotifyListenerNoCoverPayloadAvailable)
> 
> def unload():
>     if self.dbNotifyListenerNoCover.isRunning():
>         self.dbNotifyListenerNoCover.stopListener()
> 
> def _wastewater_structure_layer_available_changed(self, available):
>     if available:
>         self.dbNotifyListenerNoCover.startListener()
>     else:
>         self.dbNotifyListenerNoCover.stopListener()
> 
> def dbNotifyListenerNoCoverPayloadAvailable(self, payload: str):
>     self.iface.messageBar().pushMessage(
>         "Warning",
>         payload,
>         level=Qgis.Warning,
>         )
> ```

Thanks for the code example, really helpful. I will follow the idea on lower priority for database-wide warnings.

Issue #329, opened 2024-07-20
12.1_roles.sql error


sjib:
**Describe the bug**

When trying to run 12.1_roles.sql on a database called `teksi_pre_2024.0` the following syntax error occurs.
```
ERROR:  Syntaxfehler bei ».0«
LINE 1: GRANT CREATE ON DATABASE teksi_pre_2024.0 TO "tww_user"
                                               ^
QUERY:  GRANT CREATE ON DATABASE teksi_pre_2024.0 TO "tww_user"
CONTEXT:  PL/pgSQL-Funktion inline_code_block Zeile 1 bei EXECUTE 

FEHLER:  Syntaxfehler bei ».0«
SQL state: 42601
```

**To Reproduce**
Exact steps to reproduce the behavior:
![20240720_roles_12 1_error](https://github.com/user-attachments/assets/4df23450-b36d-425d-a6e1-110e3e4c65a1)

**Expected behavior**
_A clear and concise description of what you expected to happen._

No error should appear.

**Desktop (please complete the following information):**
 - TWW version 2024.0 pre-release
 - QGIS Version 3.34.8
 - OS Windows 10

**Additional context**

sjib:
When I quote the database name it works
`
DO $$ BEGIN EXECUTE 'GRANT CREATE ON DATABASE ' || '"' || (SELECT current_database()) || '"' || ' TO "tww_user"'; END $$;  -- required for ili2pg imports/exports`

cymed:
I strongly discourage to include points in any postgres name to avoid using quotes everywhere

ponceta:
I would avoid setting quotes and dots into database names.

https://stackoverflow.com/questions/2878248/postgresql-naming-conventions

There is no written convention since it is always possible to bypass it technically but I would avoid going offroad and recommand a lowercase with underscores naming in postgreSQL.

ponceta:
https://github.com/teksi/wastewater/pull/402

PR #330, opened 2024-07-20
 12.1_roles.sql error - quote database name

sjib:
Should solve https://github.com/teksi/wastewater/issues/329


cymed (approved):
Works, but becomes obsolete with #120

ponceta:
I would avoid setting quotes and dots into database names.

https://stackoverflow.com/questions/2878248/postgresql-naming-conventions

There is no written convention since it is always possible to bypass it technically but I would avoid going offroad and recommand a lowercase with underscores naming in postgreSQL.

ponceta:
Workaround, document postgreSQL naming convention :

https://github.com/teksi/wastewater/pull/402

PR #333, opened 2024-07-23
Do not insert all oid prefixes on db creation level

cymed:
## General
 - [x] Fix a bug

### Context
It is not a good practice to store the oid prefixes in the main datamodel. The sysadmin should add their own prefixes to their datbase instead

## Describe your changes
Remove provided oid prefixes

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [ ] CI Tests are green
- [ ] My work is ready for review

## Checklist before merge
- [ ] Comments are resolved
- [ ] Documentation is ready

### DIFF
--- a/datamodel/changelogs/0001/oid_generation.sql
+++ b/datamodel/changelogs/0001/oid_generation.sql
@@ -19,15 +19,7 @@ COMMENT ON TABLE tww_sys.oid_prefixes
 
 -- sample entry for Invalid - you need to adapt this entry later for your own organization
 INSERT INTO tww_sys.oid_prefixes (prefix,organization,active) VALUES ('ch000000','Invalid',TRUE);
-INSERT INTO tww_sys.oid_prefixes (prefix,organization,active) VALUES ('ch11h8mw','Stadt Uster',FALSE);
-INSERT INTO tww_sys.oid_prefixes (prefix,organization,active) VALUES ('ch15z36d','SIGE',FALSE);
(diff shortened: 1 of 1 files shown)


ponceta:
I feel we should ease as much the configuration of things at least for TEKSI users and why not preconfigure oid prefixes?
We discussed about having an oid prefix switch in the TMMT so having several sample oid prefixes is a nice thing at least for testing purposes.

ponceta:
In TMMT, I would like to be able to set the oid prefix from a dropdown / search menu :

image

Being a `TEKSI member` would enable you to have your `oid_prefix` already in that list without having to do sql or pgAdmin stuff.

cymed:
> In TMMT, I would like to be able to set the oid prefix from a dropdown / search menu :
> 
> image
> 
> Being a `TEKSI member` would enable you to have your `oid_prefix` already in that list without having to do sql or pgAdmin stuff.


I think it is better to have a GUI allowing to set that value instead. You'll have to do it once and nobody else needs that information.

sjib (approved):
Ok for me to take this out here. With more than 40 members and xxx communities this would become too big and difficult to maintain.

ponceta:
If we want to have INTERLIS checks in the future, we have at least to have an invalid and a valid prefix in the sample data.

cymed:
@ponceta we can use the prefix `ch24eax1`, I created it for the TEKSI2AG-xx gateway. However, we should only import it for tests, not as part of creating od

cymed:
@ponceta are you ok with merging this PR? Your comment is separate from it

PR #335, opened 2024-07-26
2024-07-26-adapt-documentation-interlis-import-export-etc

sjib:
Updating documentation to tww 2024.0

- [x] new version of toolbar image added
- [x] new INTERLIS button images
- [x] some small corrections and precisions
- [x] correct image names
- [x] update default_values

PR #338, opened 2024-07-26
Revert "Add extension-guide to transifex config"

cymed:
Reverts teksi/wastewater#332 to fix #335 problems

Adding new strings to transifex does not seem to be automatic


sjib (approved):
take back

Issue #337, opened 2024-07-26
INTERLIS Export - subclass_check


sjib:
**Describe the bug**
When exporting data with INTERLIS and the _subclass_check throws and error to export process is not stopped anymore and the ERROR only shows up in the depth of the tww protocol:

`2024-07-26T10:29:03        teksi_wastewater.utils:ERROR: number of subclass elements of structure_part NOT CORRECT in schema tww_od: checksum = 1 (positive number means missing entries, negative means too many subclass entries)`

![20240726_subclass_check_missing_structure_part](https://github.com/user-attachments/assets/f8ebe2cd-d67c-4363-9610-d380a20b66c9)

**To Reproduce**
Exact steps to reproduce the behavior:
1. Create additional data entry in superclass e.g. structure_part (that has no reference subclass entry)
![20240726_subclass_check_missing_add_extra_structure_part](https://github.com/user-attachments/assets/f5330689-9f5a-422b-956a-35f7d138d014)

3. Click on 'INTERLIS export'
4. Select export file and start export
5. See valid export message
6. Check for `ERROR: number of subclass elements of` in protocol tab 'tww'
7. Find entry

**Expected behavior**
_A clear and concise description of what you expected to happen._

When subclass_check throws an error, the export process should be canceled and a red error message should appear message bar.

**Desktop (please complete the following information):**
 - TWW version pre-release 2024.0
 - QGIS Version 3.34.8
 - OS windows 10

**Additional context**

sjib:
ilivalidator does not find this error and skips the export of this extra dataset - but the user will not be aware of this because he/she gets a success message.

sjib:
Adapt error handlinge here in `_check_subclass_count`:
https://github.com/teksi/wastewater/blob/cebe6180519a0819d71d1e2ff812afaae3b8c4bf/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py#L557

sjib:
@cymed Any quick idea how to improve this? How can we exit the export process in tww?

This is how it is implemented in qgep:
https://github.com/QGEP/qgepqwat2ili/blob/48c8a7fe15f55a26c8ba0c65eb43eee23573cb55/qgepqwat2ili/gui/__init__.py#L424

PR #340, opened 2024-07-26
raise error on subclass check

cymed:
Solves #337 
Not tested yet - thus draft PR

### DIFF
--- a/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
+++ b/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
@@ -474,7 +474,7 @@ class InterlisImporterExporter:
                 log_path,
             )
 
-    def _check_subclass_counts(self, limit_to_selection):
+    def _check_subclass_counts(self, limit_to_selection=False):
         self._check_subclass_count(
             config.TWW_OD_SCHEMA,
             "wastewater_networkelement",
--- a/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
+++ b/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
@@ -531,7 +531,6 @@ class InterlisImporterExporter:
         )
 
     def _check_subclass_count(self, schema_name, parent_name, child_list):
-
         logger.info(f"INTEGRITY CHECK {parent_name} subclass data...")
         logger.info("CONNECTING TO DATABASE...")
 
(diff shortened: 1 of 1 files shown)


plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -554,8 +553,15 @@ def _check_subclass_count(self, schema_name, parent_name, child_list):
                         f"OK: number of subclass elements of class {parent_name} OK in schema {schema_name}!"
                     )
                 else:
-                    logger.error(
-                        f"ERROR: number of subclass elements of {parent_name} NOT CORRECT in schema {schema_name}: checksum = {parent_count} (positive number means missing entries, negative means too many subclass entries)"
+                    if parent_count > 0:
+                        errormsg = f"Too many subclass entries for {schema_name}.{parent_name}"
```
sjib, line 557: It has to be the other way round: Too few here

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -554,8 +553,15 @@ def _check_subclass_count(self, schema_name, parent_name, child_list):
...
+                    else:
+                        errormsg = f"Too few subclass entries for {schema_name}.{parent_name}"
```
sjib, line 559: Too many

sjib:
@cymed Fix seems to work - only error message is wrong (change too many / too few). I have added one additional structure_part entry in my data. So I am missing a corresponding subclass entry.

20240726_subclass_check_missing_structure_part_test_fix

domi4484 (approved):

sjib:
@cymed 
If I have one entry too much in the superclass (structure_part) I get this message:
`2024-07-30T15:42:02        teksi_wastewater.utils:Subclass Count error: Too many subclass entries for tww_od.structure_part`

Should be adapted to
`2024-07-30T15:42:02        teksi_wastewater.utils:Subclass Count error: Too many superclass entries for tww_od.structure_part`

And the other case: `Too few subclass entries for {schema_name}.{parent_name}`

Is that clearer?

cymed:
Wouldn't it be too many subclass/superclass entries?

sjib:
Yes, I think then it is right - either we have too many superclass entries or we have too many subclass entries.

sjib:
And in using superclass and subclass we indicate where to check

sjib:
I found out, that we also do the subclass check when we make a selection. 

20240730_export_selection_subclass_check_error_sinnvoll

For the user it is a little bit confusing to get the subclass check error also then even if the selected objects are correct. 

- [ ] do we want to keep the subclass check also there to enforce data integrity or 
- [ ] take it out then (and then might have strange errors)?

I would stay with the existing behaviour an may be adjust the error message to not confuse the users.

Any comments on that?

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -554,9 +569,21 @@ def _check_subclass_count(self, schema_name, parent_name, child_list):
...
                 else:
-                    logger.error(
-                        f"ERROR: number of subclass elements of {parent_name} NOT CORRECT in schema {schema_name}: checksum = {parent_count} (positive number means missing entries, negative means too many subclass entries)"
-                    )
+                    if parent_count > 0:
+                        errormsg = f"Too many superclass entries for {schema_name}.{parent_name}"
+                    else:
+                        errormsg = f"Too many subclass entries for {schema_name}.{parent_name}"
+                    if limit_to_selection:
+                        logger.warning(
```
domi4484, line 577: I would add one empty line for readability :) Otherwise good for me

```suggestion
                    else:
                        errormsg = f"Too many subclass entries for {schema_name}.{parent_name}"

                    if limit_to_selection:
                        logger.warning(
```

domi4484 (approved):

sjib (approved):

Issue #341, opened 2024-07-26
Adding second value to default_values fails


sjib:
**Describe the bug**
_A clear and concise description of what the bug is._

When trying to add a second value to the default_values table I get a not NULL constraint error

**To Reproduce**
Exact steps to reproduce the behavior:
1. Go to layer group  'configuration'
2. Click on 'default_values' and open attribute table
3. Click edit mode
4. Add new object
5. See error

**Expected behavior**

It should be possible to add more than one entry

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

![20240726_adding_second_entry_default_values](https://github.com/user-attachments/assets/c5e43acc-c063-4430-ace6-d298a0f082e1)

**Desktop (please complete the following information):**
 - TWW version pre-release 20204.0
 - QGIS Version 3.34.8
 - OS Windows 10

**Additional context**

sjib:
If I add an additional data set with pg admin it works
![20240726_adding_extra_entry_default_values_pg_admin](https://github.com/user-attachments/assets/b1ca6a35-c696-40bd-bf3e-a35157044527)

Interestingly the value list shows list values without brakets and the new value with brakets.

sjib:
@urskaufmann Can you have a look at this?

sjib:
Shouldn't we provide the four basic fk in the list of the form 

- [ ] fk_owner (Eigentuemer)
- [ ] fk_provider (Betreiber)
- [ ] fk_dataowner (Datenherr) and
- [ ] fk_dataprovider (Datenlieferant)

sjib:
@cymed Or is this defined in the sql?

https://github.com/teksi/wastewater/blob/main/datamodel/changelogs/0001/14_default_values.sql

cymed:
The problem is that we have no default value to set and we evaluate defaults on the provider side. I will push a fix

sjib:
On startup I now get this error (still with the old version):
`2024-07-26T14:56:16     CRITICAL    Error : Could not check database validity: tuple indices must be integers or slices, not str`

sjib:
![20240726_form_default_values](https://github.com/user-attachments/assets/fed902ac-a490-431b-89bc-8dec1519f49f)

sjib:
- [x] Take out NOT null in data model see https://github.com/teksi/wastewater/pull/342/commits/5ff4b738dfedc868e56d0c2ab6141cf408115b0e
[...]

PR #342, opened 2024-07-26
introduce serial pk on default values

cymed:
solves #341

### DIFF
--- a/datamodel/app/view/set_default_value_for_views.sql
+++ b/datamodel/app/view/set_default_value_for_views.sql
@@ -12,7 +12,7 @@ BEGIN
 	LEFT JOIN information_schema.tables t
 	ON c.table_name = t.table_name
       AND c.table_schema = t.table_schema
-    WHERE c.column_name IN ('fk_provider','fk_dataowner')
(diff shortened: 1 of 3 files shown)


sjib:
20240726_form_default_values_corrected_table

Table now works with second and third value

sjib:
@cymed With you're new patch applied I get now the following error: `CRITICAL    Error : Could not check database validity: name 'item' is not defined`

20240801_validity_check_on_startup
20240801_validity_check_tww_od default_values
20240801_validity_check

I hope I did all well.

Maybe still something to adapt here?

    @staticmethod
    def get_validity_check_issues() -> List[str]:
        messages = []
        messages = DatabaseUtils.check_oid_prefix()
        messages.extend(DatabaseUtils.check_fk_defaults())

        if not DatabaseUtils.check_symbology_triggers_enabled():
            messages.append("Symbology triggers are disabled")

        return messages

Else I have to make another test on a new clean version.

cymed:
Should work after renaming the iterator

sjib:
When applying the new datamodel/changelogs/0001/14_default_values.sql without fk_dataprovider it works for me.
thanks @cymed

sjib (approved):
Looks ok from my side.

PR #358, opened 2024-08-06
add Linking Relation Editor and Document Management System to plugin …

cymed:
…dependencies

We use both the LRE and the DMS in our forms. By adding them as a dependency, we ensure they are loaded


ponceta:
CI looks to be broken due to external reasons (ubuntu docker image)

ponceta (approved):

PR #363, opened 2024-08-07
Remove obsolete v1 docker-compose syntax

ponceta:
Remove obsolete v1 docker-compose syntax


ponceta:
@cymed @sjib this will fix the CI issue.

Issue #365, opened 2024-08-07
INTERLIS Import: Data in re_tables is ducplicated when re-importing


sjib:
**Describe the bug**
_A clear and concise description of what the bug is._

When re-importing the same dataset or another dataset in a existing database with data it does not check if there is already the same data in the re_tables. This leads to duplicates in re_maintenance_event_wastewater_structure and re_building_group_disposal.

**To Reproduce**
Re-import a dataset with re_data a second time and check your re_table in postgres

**Expected behavior**

On import it should be checked as with all other tables if there is already a dataset with the same combination fk_1 / fk_2.
The function `create_or_update` should be enhanced or an new function should be written that checks on the two fk_* attributes of a n:m re_table instead of the obj_id if there is already the same dataset in there.

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

**Desktop (please complete the following information):**
 - TWW version https://github.com/teksi/wastewater/pull/357
 - QGIS Version 3.38.4
 - OS Windows 10

**Additional context**
_Add any other context about the problem here._

Tests with `create_re_class_entry `do not work as expected - moved to this new issue

- [ ] How to get the classname out of variable cls?
 teksi_wastewater.utils:Not supported n:m relation class: - please open an issue to add this to interlis_importer_to_intermediate_schema configuration

So it does not find the cls

- [ ] How to make this function more flexible in not having to hardcode the supported re_* classes?

```
    def create_re_class_entry(self, cls, **kwargs):
        """
        Checks if an existing instance (if fk_1, fk_2 combination is found) or creates an instance of the provided re_class
        with given kwargs, and returns it.
        """
        instance = None

        # We try to get the instance from the session/database
        # does not work as we can then not run instance = cls(**kwargs) to create object at the end
        # fk_1_value = kwargs.get("fk_1", None)
        # fk_2_value = kwargs.get("fk_2", None)

        logger.info(
                    f"Not supported n:m relation class: {cls} - please open an issue to add this to interlis_importer_to_intermediate_schema configuration"
                )

        if cls == "re_maintenance_event_wastewater_structure":
            fk_1_value = kwargs.get("fk_wastewater_structure", None)
            fk_2_value = kwargs.get("fk_maintenance_event", None)
        elif cls == "re_building_group_disposal":
            fk_1_value = kwargs.get("fk_disposal", None)
            fk_2_value = kwargs.get("fk_building_group", None)
        else:
            fk_1_value = None
            fk_2_value = None

[...]

PR #357, opened 2024-08-05
2024 08 05 import export re maintenance event

sjib:
## General
 - [x] Fix a bug

## Describe your changes
Should solve https://github.com/teksi/wastewater/issues/352 
- [x] missing re_maintenance_event data on import and export

## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [x] CI Tests are green
- [ ] The documentation is up to date with the proposed change.
- [ ] My work is ready for review

## Checklist before merge
- [ ] A review has been performed
- [ ] Comments are resolved
- [ ] Documentation is ready

PR #369, opened 2024-08-07
2024-08-07-unique-constraint-fk_1-fk_2-to-re_-tables

sjib:
Should solve https://github.com/teksi/wastewater/issues/365


ponceta:
fk_building_group is not yet created, it has to be created before this additionnal statement otherwise it will fail as it is. (see docker logs in create dumps action to see the error)

https://github.com/teksi/wastewater/blob/f13c17a9919418ab49531d2bd97b5742fddf9cb1/datamodel/changelogs/0001/03_tww_db_dss.sql#L2966C58-L2966C76

sjib:
Needs https://github.com/teksi/wastewater/pull/357 to be user friendly, else we will get a problem during import as duplicates will be blocked. This would be ok as long as we re-import the same re_table data, but makes it difficult if there are some new ones and some old one.

sjib:
@ponceta l moved unique constraints to the end - now it looks ok

sjib:
@cymed also ok for you?

cymed (approved):

domi4484 (approved):
Looks good to me

ponceta (approved):

PR #377, opened 2024-08-13
Update agxx docs

cymed:
- Pictures with updated UI
- improved description of extension


sjib (commented):
In order to follow these limitations, there is an additional foreign key on ``tww_od.cover`` pointing to ``tww_od.wastewater_node``.

@cymed How do you call this fk_cover as it is an ag64_96 extension? Do we have a notation for that already in the developers guide?

cymed:
> In order to follow these limitations, there is an additional foreign key on `tww_od.cover` pointing to `tww_od.wastewater_node`.
> 
> @cymed How do you call this fk_cover as it is an ag64_96 extension? Do we have a notation for that already in the developers guide?

ag64_fk_wastwater_node (it is a fk on tww_od.cover)

cymed:
Alterations to project (adding a ag-xx view for the corresponding forms) will be included via tww2ag6496

sjib (approved):
Let's start like this. And then evolve. Rather document too much than too little.

Issue #368, opened 2024-08-07
INTERLIS Import last_modification is updated also with append


sjib:
**Describe the bug**
_A clear and concise description of what the bug is._

When re-importing the same dataset e.g. organisation it shows in the wizard, that it already exists and only appends new data. (append mode).

The strange thing is, that also for the existing objects in this append mode, the last_modification field is updated.

**Expected behavior**
_A clear and concise description of what you expected to happen._

In append mode (as currently implemented) existing datas last_modification field should not be updated to the time of a re-import.
This behaviour should be implemented in a update mode, when an existing dataset is really updated and e.g. an attribute is changed.

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

see https://github.com/teksi/wastewater/pull/367#issuecomment-2274274925

**Desktop (please complete the following information):**
 - TWW version https://github.com/teksi/wastewater/pull/367
 - QGIS Version 3.34.8
 - OS Windows 10

**Additional context**
_Add any other context about the problem here._

To be fixed with TIT project

sjib:
![20240806_import_export_change_of_lastmodification_xtf](https://github.com/user-attachments/assets/1dbf7210-d359-49d1-97d9-453434430da3)

If we import a dataset and re-import it the last_modification is changed.

Is this a behaviour we want or not? @cymed @teksi/technical-group

cymed:
solved with #381

PR #381, opened 2024-08-15
harmonize trigger enabling/disabling

cymed:
## General
 - [x] Fix a bug
 - [x] Add a feature
 - [x] Maintenance / sustainability
 - [x] Add Documentation

## Describe your changes

We need a way to enable/disable last_modification triggers from the plugin, see #368. 

This PR creates the corresponding DB functions and calls them before importing.

Also, it moves symbology trigger checks out of the app folder, as they are created in sys

- [x] CI Tests are green

## Checklist before merge
- [ ] All required reviews have been performed
- [x] Comments are resolved
- [x] Documentation is ready


cymed:
@ponceta is there a flag to ignore tww_app in changelogs? In this use case (find triggers pointing to tww_app) it makes sense to not have it in sys

cymed:
@ponceta is there a reason why the last modification trigger is in the sys schema? We could move it to app imho

ponceta:
Many things have been changed and it is not fully clear to me why :

1. Drop and create symbology trigger have been removed for unclear reasons :
SELECT tww_sys.create_symbology_triggers()
SELECT tww_sys.drop_symbology_triggers()

They are still in the docs :
https://github.com/teksi/wastewater/blob/a1c7f464f6f011ee46b0edce8cf8454a75ad89bc/docs/en/user-guide/How-To/index.rst#L93

2. I share the thoughts that sys functions should not be in the changelogs (thought for datamodel deltas and not application modification)

3. _app vs _sys :
_app schema should be easily "droppable" and recreated.
How to upgrade
_sys schema is yet not clear to me.

I think a clear vision on how we want it with @ponceta @cymed and @3nids would be required to go further on that path.

cymed:
> Many things have been changed and it is not fully clear to me why :
> 
>     1. Drop and create symbology trigger have been removed for unclear reasons :
>        SELECT tww_sys.create_symbology_triggers()
>        SELECT tww_sys.drop_symbology_triggers()
Dropping and creating triggers requires more rights than enabling/disabling, and it is easier to check whether the triggers are active or not.
> 
> 
> They are still in the docs :
> 
> https://github.com/teksi/wastewater/blob/a1c7f464f6f011ee46b0edce8cf8454a75ad89bc/docs/en/user-guide/How-To/index.rst#L93


docs are altered in this PR.
> 
>     2. I share the thoughts that sys functions should not be in the changelogs (thought for datamodel deltas and not application modification)
sys functions should be in the changelogs, but not in the app folder

> 
>     3. _app vs _sys :
>        _app schema should be easily "droppable" and recreated.
>        How to upgrade
>        _sys schema is yet not clear to me.
Once we know how to upgrade od we use the same mechanism for sys

> 
> I think a clear vision on how we want it with @ponceta @cymed and @3nids would be required to go further on that path.

datamodel/changelogs/0001/02_sys.sql
```diff
@@ -21,3 +21,120 @@ BEGIN
...
+		AND t.tgenabled = 'D';
+  IF _disabled_count=0 THEN
+    return true;
+  ELSE
+    return false;
+  END IF;
+END;
+$DO$ LANGUAGE plpgsql SECURITY DEFINER;
+
+
```
ponceta, line 84: I'm no great fan of auto_update term, can we define some better naming for this?

We talk about triggers that set a new modification date and user to any change in the database right?

I would talk about `system_trigger` or `modification_trigger`. The fact that it is automated is in my understanding a bit implicit since it is a trigger.

cymed:
Decisions by @3nids, @ponceta  and @cymed:

- keywords "symbology_" and "modification_"
- Last modification triggers are moved to app
- Enabling/disabling triggers is to be transferred into python logic (as in create_app)

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -132,13 +132,13 @@ def interlis_import(self, xtf_file_input, show_selection_dialog=False, logs_next
             self._import_manage_organisations()
 
             # Reenable symbology triggers
-            self._progress_done(97, "Reenable symbology triggers...")
-            self._import_enable_symbology_triggers()
+            self._progress_done(97, "Reenable symbology and  modifiation triggers...")
```
ponceta, line 135:
```suggestion
            self._progress_done(97, "Reenable symbology and modification triggers...")
```

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
```diff
@@ -97,8 +97,8 @@ def interlis_import(self, xtf_file_input, show_selection_dialog=False, logs_next
         self._import_xtf_file(xtf_file_input=xtf_file_input)
 
         # Disable symbology triggers
-        self._progress_done(35, "Disable symbolgy triggers...")
-        self._import_disable_symbology_triggers()
+        self._progress_done(35, "Disable symbology and modification triggers...")
+        self._import_disable_symbology_and_modification_triggers()
```
ponceta, line 101: If you deactivate your modification triggers, you won't be able to tell what were the entities modified by your import. I would suggest to deactivate only symbology triggers.

datamodel/app/symbology_functions.sql
```diff
@@ -1,3 +1,132 @@
+------ LAST MODIFIED -----
+CREATE FUNCTION tww_app.modification_last_modified() RETURNS trigger AS $$
```
ponceta, line 2: Can you please move it to a new `datamodel/app/modification_functions.sql`

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
cymed: not possible: In AG-64/96, the inserts/updates are handled by triggers, not deactivating the modification triggers causes psycopg problems due to pending trigger events.
Also, we need to disable the last_modification triggers because otherwise we won't be able to import the base data.

datamodel/app/symbology_functions.sql
```diff
@@ -1,3 +1,132 @@
...
+			EXECUTE FORMAT('ALTER TABLE %I.%I %s TRIGGER %I',schdf,tbldf,upper(action_name),trig);
+		END LOOP;
+
+	RETURN;
+END IF;
+END;
+$DO$
+LANGUAGE plpgsql SECURITY DEFINER;
+
+CREATE OR REPLACE FUNCTION tww_app.check_modification_enabled (action_name text) RETURNS BOOL AS
```
ponceta, line 112:
```suggestion
CREATE OR REPLACE FUNCTION tww_app.check_modification_enabled () RETURNS BOOL AS
```

datamodel/app/symbology_functions.sql
```diff
@@ -1,3 +1,132 @@
...
+		INNER JOIN pg_proc p on t.tgfoid=p.oid
+		WHERE p.proname  LIKE 'symbology_%'
+		and p.pronamespace::regnamespace::text LIKE 'tww_ap_'
+		AND t.tgenabled = 'D';
+  RETURN _disabled_count=0;
+END;
+$DO$ LANGUAGE plpgsql SECURITY DEFINER;
+
+
+CREATE OR REPLACE FUNCTION tww_app.alter_modification_triggers (action_name text) RETURNS VOID AS
```
ponceta, line 81:
```suggestion
CREATE OR REPLACE FUNCTION tww_app.alter_modification_triggers(action_name text) RETURNS VOID AS
```

datamodel/app/symbology_functions.sql
ponceta, line 112: Move to a new datamodel/app/modification_functions.sql

ponceta:
Since we are at reorganizing, I would if possible move app functions to an `app/functions/` folder.

ponceta:
This is really a great cleanup for the future, and this will clearly ease our maintenance efforts for the next years.

plugin/teksi_wastewater/interlis/interlis_importer_exporter.py
ponceta: Ok then let's keep it that way for now and tackle it in the import/export issue 👍

ponceta:
image

I'll run some test and then we will be able to merge this! Thank you a lot!

ponceta (approved):
This is great and works fine! Thank you @cymed

sjib:
I think I have to take out these lines in the datamodel sql scripts:

`ALTER TABLE tww_od.organisation ALTER COLUMN obj_id SET DEFAULT tww_sys.generate_oid('tww_od','organisation');`

and the trigger statements

-------
```
CREATE TRIGGER
update_last_modified_sia405cc_cable_point
BEFORE UPDATE OR INSERT ON
 tww_od.sia405cc_cable_point
FOR EACH ROW EXECUTE PROCEDURE
 tww_sys.update_last_modified();

Anything else @ponceta @cymed ?
```

cymed:
> I think I have to take out these lines in the datamodel sql scripts:
> 
> `ALTER TABLE tww_od.organisation ALTER COLUMN obj_id SET DEFAULT tww_sys.generate_oid('tww_od','organisation');`
> 
> and the trigger statements
> 
> ```
> CREATE TRIGGER
> update_last_modified_sia405cc_cable_point
> BEFORE UPDATE OR INSERT ON
>  tww_od.sia405cc_cable_point
> FOR EACH ROW EXECUTE PROCEDURE
>  tww_sys.update_last_modified();
> 
> Anything else @ponceta @cymed ?
> ```

This is all on your side

cymed:
@ponceta good to merge?

PR #389, opened 2024-08-23
Alter UNIQUE index on organisation

cymed:
The unique identifier on organisation needs to include uid and organisation_type, as in the VSA organisation datasets there are two entries with identifier "Zeihen" (ch20p3q400000187 and ch20p3q400000338)


sjib (changes requested):
In INTERLIS the constraint is

    UNIQUE 
        !!@comment = "!! Neben UNIQUE OID zusätzlich auch Kombination Bezeichnung, Organisationstyp, UID (Wegleitung GEP-Daten 2020)"
        Bezeichnung, Organisationstyp, UID; 
    END Organisation;

So you should take out the fk_dataowner - organisations have to be unique over all dataowners.

sjib:
I have to adjust this in the create code also for future releases

sjib (approved):
Thanks @cymed

Issue #187, opened 2024-03-14
Feature Request: TEKSI Wastewater to AG-64/AG-96


cymed:
## General information

- Applicant:  Waldburger Ingenieure AG
- Developers: GeoWerkstatt GmbH, Waldburger Ingenieure AG
- User group referent: pending
- Developments referent: @el-wiss

## Users and administrators needs

- The canton of Aargau requires that wastewater cadastre and GEP topics are provided according to its own cantonal models AG-64 and AG-96 instead of the national data models SIA-405 or VSA DSS (mini) 2020.1. The AG-64 data scope can be derived from the AG-96 data scope
- In order for users operating in the canton of Aargau to be able to use the TEKSI wastewater solution, it must have an interface to the cantonal data models.

### Description of users and administrators feature requirements

_Describe the new feature_

#### General requirements
The objectives of the project are as follows:

- Expansion of the TEKSI Wastewater 2024.0 datamodel to include the necessary additional tables/attributes/views etc.
- Import and export functionality from plugin and cmd tool for the AG-64 and AG-96 data models
- Option to disable AG-64 and AG-96 export from showing in the plugin

##### Import

- On the plugin side: import of a validated dataset as an INTERLIS 2.3 transfer file in the AG-64 (version 2.1.0 from 21.04.2023) and AG-96 (version 2.0.0 from 21.04.2023) models into the TEKSI database via schema pg2ili Wastewater
- On the database side: Creation of trigger logic for semantic transformation from AG-96 to TEKSI Wastewater 2024.0

##### Export

- Export of a dataset from the TEKSI database into a valid INTERLIS 2.3 transfer file in the model AG-64 (wastewater cadastre) and AG-96 (GEP)
- Adaptation of the functionality for the export of labeling positions to include labels for building groups and measures
- Additional function for automatic integration of the current Aargau organization table during data export for organization table mapping when an AGxx export is executed (pre-assessments showed that it is smarter to use the VSA organisations and map them to AGxx-organisations, otherwise the VSA DSS compatibility is compromised)
- Conversion of the geometry for the export of the aggregated catchment areas of the special structures (in the data model: class SBW_Einzugsgebiet)

### Illustrations of the features and the requirements

- creation of new optional extension schema on database
- adaptation of qgis project to include building groups and measures
- Extension of plugin for INTERLIS Export

### Graphical User Interface (GUI)

- Add Checkbox in the developer options of the settings dialog to enable/disable the AG-64/96-Export
[...]

PR #399, opened 2024-09-03
Add AG-64/96 extension

cymed:
This Pull request pushes the extension for AG-64/96 to TEKSI Wastewater.
Depends on #381

Closing #187


cymed:
Error: SIA405_Base_Abwasser_1_LV95: model(s) not found

.editorconfig
ponceta, line 1: Add to .gitignore and remove from the PR

Philippluca: Just removed from PR. .gitignore only relevant when deciding to work with VS Code explicitly not.

cymed:
@Philippluca can you check the error in pre-commit.ci and whether it is linked to the other ci changes?

sjib:
@cymed When is this ready to be integrated? 2024.x?

cymed:
It is ready, but I suggest releasing it in .1 shortly after .0. This way, if we encounter unexpected problems, we have an official version to fall back to

ponceta:
This is strongly linked to what we will support with TMMT, I do agree with @cymed.

PR #409, opened 2024-09-06
separate labels from geometries

cymed:
Same idea as #85, but from scratch to not deal with heritage code


ponceta:
This is really sexy! 😎😎😎 (if wastewater trigger labels functions can be qualified in that way)

sjib:
@cymed Can you add/link the documentation about this new feature and suggest where to add it to the developers guide also.

cymed:
First I want to get it running

PR #416, opened 2024-09-12
Add specialised view for infiltration installation

cymed:
## General
 - [ ] Fix a bug
 - [x] Add a feature
 - [ ] Maintenance / sustainability
 - [ ] Add Documentation

## Describe your changes

- [x] Add specialised view for infiltration installation
- [x] includes retention_body and mechanical_pretreatment.

Context : informations: retention_body and mechanical_pretreatment are classes referencing to infiltration installation. In some cantons, it is mandatory to fill out this info. This PR creates a view that can be used to fill out the necessary forms more easily

- [x] CI Tests are green

## Checklist before merge
- [ ] Comments are resolved
- [ ] Documentation is ready


ponceta:
@cymed we need to have more context and maybe screenshots on what we want to achieve with such an evolution.

cymed:
retention_body and mechanical_pretreatment are classes referencing to infiltration installation. In some cantons, it is mandatory to fill out this info. This PR creates a view that can be used to fill out the necessary forms more easily

ponceta:
@cymed what are metafunctions and why are these included here?

cymed:
For now it is only one function that can check if all entries for a prefix are null

datamodel/app/create_app.py
```diff
@@ -9,6 +9,7 @@
 except ImportError:
     import psycopg2 as psycopg
 
+from gep_views.vw_tww_infiltration_installation import vw_tww_infiltration_installation
```
ponceta, line 12: Please move it to view.vw_tww_infiltration_installation gep_views is currently at the wrong place.

This should be reorganised in the future as :
views / vw_*
views / vw_tww_*
views / export / vw_*
views / SWMM/ vw_swmm_*
views / GWMP / vw_gwmp (gep is german) (only if we have the necessity to create some distinction)

datamodel/app/functions/meta_functions.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE FUNCTION tww_app.check_all_nulls(rec record, prefix varchar(2),ignored_postfix text[] default Array['obj_id','identifier'])
```
ponceta, line 1: Can we have a comment or some more explanation of what does this function exactly?

ponceta (changes requested):

datamodel/app/functions/meta_functions.sql
cymed: It allows checking a record (usually NEW) on whether or not its attributes with a certain prefix are all NULL. The third argument allows whitelisting auto-populated values

datamodel/app/create_app.py
cymed: resolved in https://github.com/teksi/wastewater/pull/416/commits/91d6336daa466defef9a4acdfbb789ec52468b95

ponceta (approved):

Issue #417, opened 2024-09-17
Doc / User-guide is translated to german to 99%, but does not appear (https://teksi.github.io/wastewater/de/  or where is the tww-docu, not the "dev" version).


ponceta:
Doc / User-guide is translated to german to 99%, but does not appear (https://teksi.github.io/wastewater/de/  or where is the tww-docu, not the "dev" version).

_Originally posted by @urskaufmann in https://github.com/teksi/wastewater/issues/398#issuecomment-2349148064_
            
We have to create an event action that rebuilds docs every night like we did on qgep docs.

ponceta:
https://github.com/QGEP/docs/blob/master/.github/workflows/release.yml#L6

ponceta:
The proposed PR does build but not deploy the docs content.

@3nids could you have a look at the conditions to make the translated docs deploy every morning?

PR #427, opened 2024-09-25
Add scheduled action to rebuild documentation

ponceta:
Should fix https://github.com/teksi/wastewater/issues/417

### DIFF
--- a/.github/workflows/docs-builder.yml
+++ b/.github/workflows/docs-builder.yml
@@ -17,6 +17,8 @@ on:
     paths:
       - docs/**
       - '.github/workflows/docs-builder.yml'
+  schedule:
+    - cron: "22 11 * * *"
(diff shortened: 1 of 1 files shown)


3nids (approved):
rebuilding every day sounds an overkill, no?
I would rather suggest once a week or so, and potentially add a manual trigger in case you want to see more rapidly changes https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

But that's up to you to decide.

ponceta:
@3nids Thank you for your feedback, for this time with many contributions I will set it daily as it is yet configured for qgep docs.
The manual trigger is available too.
I note to evaluate it on a weekly basis and document who is able and who to contact to run it manually.

Issue #393, opened 2024-08-29
Plugin sometimes ignores SSL certificate from pg_service.conf


cymed:
**Describe the bug**
Parts of TEKSI Wastewater's functionality do not work when connecting through a pg_service that includes ssl definition

**To Reproduce**
Exact steps to reproduce the behavior:
1. Create working pg_service entry in pg_service.conf that requires a ssl certification

[pg_tww]
sslrootcert=path_to_rootrcert/foo.crt
sslcert=path_to_sslcert/bar.crt
sslkey=path_to_sslkey/foobar.key
sslmode=verify-full
host=1.2.3.4
port=5432
dbname=pg_tww

2. Start the TEKSI project. 
3. Test functionalities:

+ Upstream works
+ Downstream works
+ Opening/closing wizard works
+ Altering geometries works
+ Connecting wastewater networkelements works

- Interlis Import causes "FATAL: connection requires a valid client certificate"
- Interlis Export causes "FATAL: connection requires a valid client certificate"
- Update Symbology causes "FATAL: connection requires a valid client certificate"
- Check Database Validity causes "FATAL: connection requires a valid client certificate"
- Enable Symbology Triggers causes "FATAL: connection requires a valid client certificate"
- Disable Symbology Triggers causes "FATAL: connection requires a valid client certificate"

**Expected behavior**
Plugin passes SSL client certificate for all functionalities if it is defined in the pg_service.conf

**Desktop (please complete the following information):**
 - TWW version: 2024.0_pr
 - QGIS Version: 3.34.8
 - OS: Windows

**Additional context**
I suppose the ssl values are not passed in the database utils.

ponceta:
Thank you @cymed for this pertinent feedback. We have to take this into account when rewriting some of these components.

Can you tell me if the edition of such parameters is already available in pg service parser plugin :

https://plugins.qgis.org/plugins/pg_service_parser/

Since we aim to use its library in the future it would be good to see if any evolution is required on that side and then open a feature proposal on their repo.

cymed:
The pg_service_parser plugin currently supports host, port, dbname, user, password, sslmode and passfile afaik. 
We'll also have to take into consideration the --dbparams parameter in ili2db
TEKSI's Database Config currently passes host, port, dbname, user, password

cymed:
the pgserviceparser library seems to pass the additional parameters, but the pg_service_parser plugin does not support all of them.

ponceta:
[...]

PR #430, opened 2024-09-27
Allow SSL params for Auth

cymed:
This PR aims to fix the problems in #393. Now. all parameters in the pg_service entry are evaluated by the plugin. When using INTERLIS-Import/Export, all values that are not host/port/database/user/password are stored in a temporary dbparams file for usage in ili2pg. 

The main problem I have now is that that SSL-keys must be in DER format to be used in ili2pg, but QGIS requires PEM. Everything except INTERLIS works with SSL. See https://github.com/claeis/ili2db/issues/551 for upstream issue

### DIFF
--- a/plugin/teksi_wastewater/interlis/utils/various.py
+++ b/plugin/teksi_wastewater/interlis/utils/various.py
@@ -38,26 +38,26 @@ def get_pgconf_as_ili_args() -> List[str]:
     args = []
     dbparams = []
     for key in pgconf:
-        if key=="host":
+        if key == "host":
             args.extend(["--dbhost", '"' + pgconf["host"] + '"'])
-        elif key=="port":
+        elif key == "port":
             args.extend(["--dbport", '"' + pgconf["port"] + '"'])
-        elif key=="user":
+        elif key == "user":
             args.extend(["--dbusr", '"' + pgconf["user"] + '"'])
-        elif key=="password":
+        elif key == "password":
             args.extend(["--dbpwd", '"' + pgconf["password"] + '"'])
-        elif key=="dbname":
+        elif key == "dbname":
             args.extend(["--dbdatabase", '"' + pgconf["dbname"] + '"'])
         else:
-            dbparams.extend([f'{key}={pgconf[key]}'])
+            dbparams.extend([f"{key}={pgconf[key]}"])
     if dbparams:
         # write into tempfile and add path to args
-        dbparams_path=os.path.join(tempfile.gettempdir(), "dbparams")
+        dbparams_path = os.path.join(tempfile.gettempdir(), "dbparams")
         os.makedirs(dbparams_path, exist_ok=True)
-        with open(os.path.join(dbparams_path,"dbparams.txt"),"w") as f:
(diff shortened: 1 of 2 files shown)


cymed:
@domi4484 does DatabaseConfig accept overridden env variables from QGIS or not?

domi4484:
@cymed I am not sure what do you mean. At this place you have the possibility to pass things from the "QGIS world": https://github.com/teksi/wastewater/blob/main/plugin/teksi_wastewater/teksi_wastewater_plugin.py#L614

Otherwise the modifications here looks correct to me, is it something that still doesn't work?

cymed:
My main problem is handling of the user name. ili2pg requires a username even if the username is implicitly passed using certs. Currently, I implemented the OS user as a fallback, but there might be situations where the pg username is distinct from the OS username.
One option would be to override the PGUSER variable in QGIS, which is then read on initialisation. I do however not know how to access these.

domi4484:
> One option would be to override the PGUSER variable in QGIS, which is then read on initialisation. I do however not know how to access these.

I guess you should be able to access it like that, but this needs to be done inside teksi_wastewater_plugin.py and passed to the DBUtils class 
```
expression = QgsExpression('@PGUSER')
pguser = expression.evaluate()
```

plugin/teksi_wastewater/interlis/utils/various.py
```diff
@@ -36,16 +37,30 @@ def get_pgconf_as_ili_args() -> List[str]:
...
+            args.extend(["--dbusr", '"' + pgconf["user"] + '"'])
+        elif key == "password":
+            args.extend(["--dbpwd", '"' + pgconf["password"] + '"'])
+        elif key == "dbname":
+            args.extend(["--dbdatabase", '"' + pgconf["dbname"] + '"'])
+        else:
+            dbparams.extend([f"{key}={pgconf[key]}"])
+    if dbparams:
+        # write into tempfile and add path to args
+        dbparams_path = os.path.join(tempfile.gettempdir(), str(uuid.uuid4()))
```
domi4484, line 56: I think there is an utility function to get an unique temp folder but I can't remember the name yet.

cymed: casting a uuid to string should have the same effect I think

ponceta (approved):

ponceta:
@cymed Fine for me, ili2db remains an issue but won't be fixed here right?

Free to merge.

cymed:
sslmode works, but ili2db and QGIS require different cert types. Once thst is fixed upstream, we need no adaptations

ponceta:
> sslmode works, but ili2db and QGIS require different cert types. Once thst is fixed upstream, we need no adaptations

is fixed or has to be fixed upstream, because I saw your issue is still open on ili2pg?

cymed:
> > sslmode works, but ili2db and QGIS require different cert types. Once thst is fixed upstream, we need no adaptations
> 
> is fixed or has to be fixed upstream, because I saw your issue is still open on ili2pg?

The upstream issue stems from the type of ssl certificate only. Once it is fixed upstream, the TEKSI code is already ready.

PR #431, opened 2024-09-30
SECURITY DEFINER for default value functions

cymed:
This PR adds SECURITY DEFINER, otherwise we have insufficient priviledge problems.


ponceta:
This enables the tww_user to change active organisations and define default values.

This was a more tww_manager task until yet.

ponceta (approved):

PR #432, opened 2024-09-30
Add VS Code development environment

Philippluca:
Closes: https://github.com/wapaAdmin/tww2ag6496/issues/81

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "qgis.installationRoot": "C:\\Program Files\\QGIS 3.34.10",
-    "qgis.executable": "${config:qgis.installationRoot}\\bin\\qgis-ltr-bin.exe",
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "qgis.installationRoot": "C:\\Program Files\\QGIS 3.34.10",
+    "qgis.executable": "${config:qgis.installationRoot}\\bin\\qgis-ltr-bin.exe",
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -58,7 +58,7 @@
                 "command": "C:\\OSGeo4W\\bin\\qgis-ltr-bin.exe"
             },
             "linux": {
-                "command": "qgis-ltr"
+                "command": "qgis"
             }
         }
     ]
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -7,9 +7,11 @@
             "label": "Install dependencies",
             "type": "shell",
             "command": "pip install debugpy \"sqlalchemy~=1.3.22\" \"geoalchemy2>=0.9.0\"",
-            "options": {
-                "shell": {
-                    "executable": "${config:qgis.installationRoot}\\OSGeo4W.bat"
+            "windows": {
+                "options": {
+                    "shell": {
+                        "executable": "C:\\OSGeo4W\\OSGeo4W.bat"
+                    }
                 }
             },
             "problemMatcher": []
@@ -33,7 +35,6 @@
         {
             "label": "Launch QGIS",
             "type": "process",
-            "command": "${config:qgis.executable}",
             "args": [
                 "--project",
                 "${workspaceFolder}\\project\\teksi_wastewater.qgs"
@@ -52,6 +53,12 @@
                     "QGIS_PLUGIN_USE_DEBUGGER": "debugpy",
                     "PGSERVICEFILE": "${workspaceFolder}\\project\\pg_service.conf"
                 }
+            },
+            "windows": {
+                "command": "C:\\OSGeo4W\\bin\\qgis-ltr-bin.exe"
+            },
+            "linux": {
+                "command": "qgis-ltr"
             }
         }
     ]
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,58 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Install dependencies",
+            "type": "shell",
+            "command": "pip install debugpy \"sqlalchemy~=1.3.22\" \"geoalchemy2>=0.9.0\"",
+            "options": {
+                "shell": {
+                    "executable": "${config:qgis.installationRoot}\\OSGeo4W.bat"
+                }
+            },
+            "problemMatcher": []
+        },
+        {
+            "label": "docker-db",
+            "type": "docker-compose",
+            "dockerCompose": {
+                "up": {
+                    "detached": true,
+                    "build": false,
+                    "services": [
+                        "db"
+                    ]
+                },
+                "files": [
+                    "${workspaceFolder}/docker-compose.yml"
+                ]
+            }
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -17,3 +17,28 @@ How to start [testing](https://github.com/teksi/wastewater/discussions/72)
...
+1. Open OSGeo4W Shell and run `pip install debugpy`.
+2. Clone this repository to your local machine.
+3. Open QGIS, go to the _Environment_ section in `Settings > Options - System` and add the following custom variables:
+
+   | Apply  | Variable                 | Value                   |
+   | :----- | :----------------------- | :---------------------- |
+   | APPEND | QGIS_PLUGINPATH          | {repositoryPath}/plugin |
+   | APPEND | QGIS_PLUGIN_USE_DEBUGGER | debugpy                 |
+
+4. Install QGIS plugin _Plugin Reloader_. This will allow you to reload the plugin without restarting QGIS.
```
domi4484, line 32: This is not working at the moment if the project is already open. To be able to use the plugin reloader we need to check on load if the wastewater project is open, and in that case pass the database configuration to the DatabaseUtils class https://github.com/teksi/wastewater/blob/main/plugin/teksi_wastewater/teksi_wastewater_plugin.py#L614

Philippluca: Q: What is your proposal? 

- Add the capability to use the Plugin Reloader by checking on load if the project is open.
- Remove this comment as QGIS with the project can easily be started by VS Code.

domi4484: you can leave the comment, I added a fix for that here: #434

ponceta: Fix merged.

ponceta:
I have a general remark on embedding visual code windows path to the environment, is that not a burden for linux devs @domi4484 ?

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "qgis.installationRoot": "C:\\Program Files\\QGIS 3.34.10",
```
ponceta, line 2: When using OSGEO4W installer which is recommended by TEKSI, the path will be different.
eg : C:\\OSGeo4W\\bin\\qgis-ltr-bin.exe

project/pg_service.conf
```diff
@@ -0,0 +1,7 @@
+[pg_tww]
+# Konfiguration für TEKSI-Entwicklungsdatenbank in docker
```
ponceta, line 2: Please use english for comments
```suggestion
# Configuration for TEKSI development database in docker
```

.vscode/settings.json
Philippluca: Q: Should this recommendation be placed within the README.md as well?
The settings.json was intended as a template. User can overwrite the [settings in vscode](https://code.visualstudio.com/docs/getstarted/settings). See also comment in README.md:

> 3. Install QGIS in Version 3.34.10 or adapt `qgis.installationRoot` in ['.vscode/settings.json'.](.vscode/settings.json)

If QGIS & Shell would be found on PATH on all Plattform, the configuration would be much simpler. Alternatively, if the required version / distribution of QGIS could be specified we would have the option to write a [platform aware launch.json](https://code.visualstudio.com/docs/editor/debugging#_platformspecific-properties) without using the "hacky" settings.json.

domi4484: If we adapt .vscode/settings.json will it not always show as modified in git status? 

> If QGIS & Shell would be found on PATH on all Plattform, the configuration would be much simpler.

Maybe the needed path could be defined in system variables. If not set, just use the most common variant (porbably windows/OSGeo4w ?). 
Or providing the file as a `settings.json.default` to be renamed?

But I am not used to this kind of vscode configuration I don't now what's the best way.

domi4484:
> I have a general remark on embedding visual code windows path to the environment, is that not a burden for linux devs @domi4484 ?

I think it should not disturb if we don't use those task. It seems all a bit system/installation dependent but if it helps other I have nothing against

plugin/teksi_wastewater/teksi_wastewater_plugin.py
```diff
@@ -82,6 +83,17 @@ class TeksiWastewaterPlugin:
     profile = None
 
     def __init__(self, iface):
+        if os.environ.get("QGIS_DEBUGPY_HAS_LOADED") is None:
```
ponceta, line 86: I'm not sure including debugpy in production environment is a good practice. 
Is there a way to include it only in a dev mode only?

domi4484: There is an existing plugin setting that we can use:
```python
self.settings.value("/TWW/DeveloperMode", False, type=bool)
```

image

Philippluca: @cymed Können wir die Anpassungen noch weiter verfolgen?

.vscode/settings.json
Philippluca: I propose to use OSGeo4w as default in settings.json and adapting it to `qgis.installationRoot.windows`. If we run int in vscode the value either is correct, or the developer will get a message like "C:\OSGeo4w\qgis.bar was not found." I assume every developer will get a hint of what to do. OSGeo4w, in the optimal case, defines a link for the lates installed QGIS version to use. Corresponding linux variant can then be added as well.

The approuch should not depend on any environment variables defined from outside of VSCode as this break the whole concept of "no setup needed launch configuration".

plugin/teksi_wastewater/teksi_wastewater_plugin.py
Philippluca: I will extend the if statement to include the Developer mode setting.

.vscode/tasks.json
```diff
@@ -0,0 +1,65 @@
...
+                    "QGIS_PLUGINPATH": "${workspaceFolder}\\plugin",
+                    "QGIS_PLUGIN_USE_DEBUGGER": "debugpy",
+                    "PGSERVICEFILE": "${workspaceFolder}\\project\\pg_service.conf"
+                }
+            },
+            "windows": {
+                "command": "C:\\OSGeo4W\\bin\\qgis-ltr-bin.exe"
+            },
+            "linux": {
+                "command": "qgis-ltr"
```
domi4484, line 61:
```suggestion
                "command": "qgis"
```

I guess qgis-ltr is not available by default on most distros

Philippluca: I understood that the development and testing should be conducted against latest ltr release. Does the ltr release lack the `qgis-ltr` binding and is only available as `qgis`? Can multiple versions even coexist on linux like they are in OSGeo4w?

domi4484: > I understood that the development and testing should be conducted against latest ltr release.

Right but in fact if you install the ltr, or another release it is always installed as `/usr/bin/qgis` (at least in the distro I am using). 
In practice I usually work with a slighter newer version for convenience. Note that the pipeline tests are run against 3.34 and latest https://github.com/teksi/wastewater/blob/main/.github/workflows/plugin-test.yml#L32

> Does the ltr release lack the `qgis-ltr` binding and is only available as `qgis`? Can multiple versions even coexist on linux like they are in OSGeo4w?

It depends how it is packaged, but to get a specific qgis version you usually need to add additional repositories.
For example in OpenSUSE you can add the Geo repo and you will get the option to install the ltr but this way you have to choose ltr or latest.

To have different versions at the same time I usually build them from source and install into a custom prefix.

ponceta (approved):

PR #438, opened 2024-10-03
Move network to app

ponceta:
To avoid unecessary deltas if table extensions are required for this functionnality.


cymed:
Moving network o app is ok, but we need to alter the tests too

ponceta:
I think we will stand with the actual logic, creating tables and columns in changelogs and creating application schemes and mechanics in the app schema

PR #446, opened 2024-10-04
add a check in project TX for invalid layers

3nids:


3nids:
test locally, fails as expected.

PR #450, opened 2024-10-10
disable modification triggers when updating all symbologies

cymed:
When updating the symbologies, it is faster to disable the last_modification triggers. We do not alter the base data, but only update dependencies. Also, when adding other modification triggers such as in agxx, these need to be turned off temporarily.


sjib:
Very good point!

ponceta (approved):

PR #453, opened 2024-10-11
prioritise discharged_wastewater in symbology

cymed:
Usually, the node symbology works fine by prioritising the outgoing channel over incoming channels. However, when discharged wastewater flows into a culverted water body (eigedoltes Gewässer), it is preferrable to use the discharged wastewater usage for symbology.


ponceta:
@cymed we're ok with having removed codes to a specific list and now re-introducing codes to the same function?

cymed:
> @cymed we're ok with having removed codes to a specific list and now re-introducing codes to the same function?

I alter the code to allow making it variable

ponceta (approved):

Issue #462, opened 2024-10-16
vw_wastewater_structure is twice in the layer order and therefore twice on the map


ponceta:
**Describe the bug**
`vw_wastewater_structure` is twice in the layer order and therefore twice on the map.

I don't now how?

**To Reproduce**
Exact steps to reproduce the behavior:
1. Right click on a detail wastewater_structure
2. Open layer order
3. See error

![image](https://github.com/user-attachments/assets/6e4886ee-d294-4cfc-a82a-1302faecf8a7)

![image](https://github.com/user-attachments/assets/69b69a57-f01a-4d57-a001-55e64c5fd52e)

**Expected behavior**

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

**Desktop (please complete the following information):**
 - TWW version
 - QGIS Version
 - OS

**Additional context**

ponceta:
![image](https://github.com/user-attachments/assets/a91d740b-0ca0-4337-b379-db73b998a73c)

Issue #460, opened 2024-10-16
QGIS project : Reach points are not loaded


ponceta:
**Describe the bug**
Reach points are not loaded in the QGIS project, therefore if any corrections around these entity is require, the user has to manually add these to the project :

See related issue : https://github.com/teksi/wastewater/issues/459

**Expected behavior**
Reach points are loaded but deactivated by default in the QGIS project.

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

**Desktop (please complete the following information):**
 - TWW version
 - QGIS Version
 - OS

**Additional context**
WARNING    Could not export to the interlis schema : Invalid dataowner reference for object 'ch000000RP000012' of type 'haltungspunkt'

urskaufmann:
Reachpoint-attributs are part of vw_tww_reach. Therefore it's not necessary to add reachpoints as layer.
If the default-values for dataowner and provider work correct, then this problem should not happen.
I did not add the fk_dataowner, fk_provider field of the reachpoints to the attributform in the project, because nobody wants to add these (nonsense-)fields. I have once defined a sql to fill these attributs automatic (wastewater-structure -> networkelement -> reachpoint). The user should never have to think about these attributs...

ponceta:
Is it critical if we add this layer?

> If the default-values for dataowner and provider work correct, then this problem should not happen.

I hope we fix it soon but in the meantime it would be really practical to be able to fix issues related to that so users can start using TWW.

sjib:
@ponceta @urskaufmann

Please align your ideas and thoughts on how things are implemented right now – I think there are different conceptions and it would be good to sort these out first, before changing something!

>

urskaufmann:
![image](https://github.com/user-attachments/assets/202bc808-f924-457c-967c-62811f332419)
All reachpoint fields are in tww_vw_reach: between the two red lines in the picture
I did not add the yellow marked fields to the form, because user want/are not allowed not to edit these attributes
But if necessary its easy to add this fields.
Instead of adding a new layer its easier to config/pre-configure this fields in vw_tww_reach.

Remember: in normal manhole case, all reachpoints will be at the same place. You can not easy select the reachpoint of a specific reach. 

It's important to have a function or trigger that sets these fk_dataowner and fk_provider to the similar value of the reach. All other values make no sense (because a reachpoint only exists if the reach exists).

ponceta:
[...]

Issue #459, opened 2024-10-16
WARNING    Could not export to the interlis schema : Invalid dataowner reference for object 'ch000000CL000006' of type 'kanal'


ponceta:
**Describe the bug**
Invalid dataowner reference for object 'ch000000CL000006' of type 'kanal'

dataowner has not been set, even if it is Mandatory for export. 

Same for reach points which should be inherited directly from reaches:
Could not export to the interlis schema: Invalid dataowner reference for object 'ch000000RP000012' of type 'haltungspunkt'

**To Reproduce**
Exact steps to reproduce the behavior:
1. Create a new reach
2. Complete every field on the main form
3. Save
4. Try to export
5. See error

**Expected behavior**
Mandatory fields are visible and easy to edit

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

**Desktop (please complete the following information):**
 - TWW version
 - QGIS Version
 - OS

**Additional context**

sjib:
This would need I guess some adaptions at the project file to show which fields are MANDATORY
I would just start with which relations are MANDATORY - for the other things you will get an INTERLIS log file with all the details.

Issue #422, opened 2024-09-19
Cut reach does not work


ponceta:
**Describe the bug**

**To Reproduce**
Exact steps to reproduce the behavior:
1. Try to cut a reach with QGIS tool
2. See error

```
2024-09-19T15:36:40     CRITICAL    Couche vw_tww_reach : Erreur PostGIS lors de l'ajout d'entité : ERREUR: la valeur d'une clé dupliquée rompt la contrainte unique « in_od_reach_point_identifier »
             DETAIL: La clé « (identifier, fk_dataowner)=(HA1339E, ch080qwzPR000018) » existe déjà.
             CONTEXT: instruction SQL « INSERT INTO tww_od.reach_point (
              obj_id
              , elevation_accuracy
              , fk_dataowner
              , fk_provider
              , fk_wastewater_networkelement
              , identifier
              , last_modification
              , level
              , outlet_shape
              , pipe_closure
              , position_of_connection
              , remark
              , situation3d_geometry
              ) VALUES (
              COALESCE( NEW.rp_to_obj_id, tww_app.generate_oid('tww_od'::text, 'reach_point'::text) )
              , NEW.rp_to_elevation_accuracy
              , COALESCE(NULLIF(NEW.rp_to_fk_dataowner,''), NEW.fk_dataowner) -- fk_dataowner
              , COALESCE(NULLIF(NEW.rp_to_fk_provider,''), NEW.fk_provider) -- fk_provider
              , NEW.rp_to_fk_wastewater_networkelement
              , NEW.rp_to_identifier
              , NEW.rp_to_last_modification
              , NEW.rp_to_level
              , NEW.rp_to_outlet_shape
              , NEW.rp_to_pipe_closure
              , NEW.rp_to_position_of_connection
              , NEW.rp_to_remark
              , ST_EndPoint(NEW.progression3d_geometry) -- situation3d_geometry
              ) RETURNING obj_id »
             fonction PL/pgSQL tww_app.ft_vw_tww_reach_insert(), ligne 40 à instruction SQL
```

**Expected behavior**
Reaches can be cut with QGIS standard tools.

![image](https://github.com/user-attachments/assets/e4efe89a-84f7-43d3-8160-a11b96826137)

**Screenshots / data**
![image](https://github.com/user-attachments/assets/4d53f2b2-2042-4925-a88a-7ef736d23b03)

**Desktop (please complete the following information):**
 - TWW version
 - QGIS Version
 - OS

**Additional context**

[...]

PR #465, opened 2024-10-16
Fix several bugs on QGIS .qgs project

ponceta:
## General
 - [x] Fix several bugs

## Describe your changes
- [x] Add reach points to the project visible from scale 1:100
- [x] Set as mandatory fk_dataowner fk_provider for reach points, wastewater structures, reaches
- [x] Set rule on reach attributes to avoid rp_obj_id duplication

## Screenshots

image

## Issue ticket number and link
Fix :

- [ ] https://github.com/teksi/wastewater/issues/462
- [ ] https://github.com/teksi/wastewater/issues/460
- [ ] https://github.com/teksi/wastewater/issues/459
- [ ] https://github.com/teksi/wastewater/issues/422

- [x] CI Tests are green

## Checklist before merge
- [ ] Comments are resolved
- [ ] Documentation is ready


ponceta:
Cutting is working as expected if reaches have a correct 3D Geometry :

image

image

image

Link between reaches has to be set with the connect tool, usually a wastewater structure / wastewater node will be set on that point.

urskaufmann:
This cutting is on one side a help (I have less manual work), on the other side there is a lot of things to think about when using this cut (levels, connections, also connections of other reaches to the trimmed reach (Blindanschluss)). The wish for a complete tool remains...

urskaufmann:
3d geometry: As long as I can not export/import the 3d geometry, I will not enforce this 3d thing. 

Vision until the ili-definition supports 3d: TEKSI exports always (user to not care about) together with the xtf a exchange-file with 3d-geometries. When importing and this exchange file exist, TEKSI imports after the xtf-part also the 3d part and updates the geometries.

cymed:
General question: Why do we need the visualisation of the reach points in the project? We can access the data via vw_tww_reach.

ponceta:
> General question: Why do we need the visualisation of the reach points in the project? We can access the data via vw_tww_reach.

- [ ] Actually only for debugging purpose now. I'll draw it back from this PR so we can still add it manually when needed.

It could be nice to display visually the outlet shapes and altitudes / in the future to edit the reach start or end point.

ponceta:
Reach point layer has been removed.

PR #470, opened 2024-10-23
Add sslmode to QGIS .qgs project

ponceta:
Add sslmode to QGIS .qgs project

- [x] sslmode=prefer as default parameter

### DIFF
--- a/.github/workflows/project-translation.yml
+++ b/.github/workflows/project-translation.yml
@@ -11,6 +11,14 @@ on:
     paths:
       - project/**
       - '.github/workflows/project-translation.yml'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - project/**
+      - '.github/workflows/project-translation.yml'
+  schedule:
+    - cron: "30 04 * * *"
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


urskaufmann:
I have always used ssl mode = prefer, when adding layers to the project. But in the project was no ssl entry. Why?

ponceta:
> I have always used ssl mode = prefer, when adding layers to the project. But in the project was no ssl entry. Why?

Probably because we use a pg_conf that can handle it directly? No clear idea why exactly, and not sure if enforcing it in the qgs file is a good idea or not. This was mainly harmonization work between TWW and TDH.

cymed:
I am in favour of leaving it empty. See also https://github.com/qgis/QGIS/issues/58936

cymed:
Also it is smart imho to only have on open pr altering the qgs project. It is easy to overwrite changes otherwise

ponceta:
As discussed in upstream QGIS issue, it makes no sense to add sslmode to a pg_service configured QGIS project.

This has to be taken into account on TDH project as well.

PR #477, opened 2024-10-28
2024 10 28 add protection tube control cable

sjib:


cymed:
Why fk_channel? fk_channel is not in the 2015 ili

PR #487, opened 2024-11-01
Add pre-export checks from qgep

sjib:


sjib:
- [ ] testing needed
- [ ] add skip_wwtp_structure_ids to matching files

Issue #485, opened 2024-10-31
deleting reaches: disabling symbology triggers leaves dead data in the database (reach_points etc)


urskaufmann:
**Describe the bug**
- open a new database prerelease 2024.03
- import xtf organisation and data
Info:    1611 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Abwasserbauwerk_Text
Info:    2273 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Abwasserknoten
Info:    1535 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Deckel
Info:      90 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Einleitstelle
Info:     126 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Einstiegshilfe
Info:    2163 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung
Info:    1519 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung_Text
Info:    4326 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Haltungspunkt
Info:    2073 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Kanal
Info:    1809 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Normschacht
Info:       9 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Rohrprofil
Info:      25 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Spezialbauwerk
Info:       2 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Streichwehr
Info:       2 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Trockenwetterfallrohr
Info:      12 objects in CLASS DSS_2020_1_LV95.Siedlungsentwaesserung.Unterhalt

- Watch this data in TWW
- Important: disable symbology triggers (otherways it lasts hours)

- delete all records in vw_tww_wastewater_structure (1924)
- delete all records in vw_tww_reach (2163)
- delete all records in vw_wastewater_node (371)

-> no more visible records in TWW

- controle with pgAdmin, how many records are still in tww_od-tables:
- wastewater_structure: 2073
- channel: 2073
- reach: 0
[...]

PR #491, opened 2024-11-11
Update functions and triggers

ponceta:
## General
 - [x] Fix a bug
 - [ ] Add a feature
 - [x] Maintenance / sustainability
 - [ ] Add Documentation

## Describe your changes

This aims to fix https://github.com/teksi/wastewater/issues/485

- [x] Rename symbology_on_reach_delete into modification_on_reach_delete
- [ ] Reorganize files in application (draft)

## Documentation

**Management functions :** 
- alter_symbology_triggers(enable/disable)
- checksymbology_triggers_enabled()
- tww_app.alter_modification_triggers(enable/disable)
- tww_app.check_modification_enabled()

**Modification functions:**
- tww_app.modification_on_reach_delete() (only on reach deletion)
-- Remove linked network element and reach points
-- Remove channel if no reach is left

**Symbology functions:**
 - tww_app.update_wastewater_node_symbology()
   - Updates selected nodes or all nodes, can be launch manually over the whole database or is triggered by 
     - a `channel` update, 
     - a `reach point` update, 
     - a `reach` update

## Issue ticket number and link
https://github.com/teksi/wastewater/issues/485

- [x] CI Tests are green

## Checklist before merge
- [x] A review has been performed
- [x] Comments are resolved
- [x] Documentation is ready


ponceta:
@domi4484 @cymed @3nids could you review it :

- [ ] From the code organisation perspective (good / bad / what to adapt?)
- [ ] From the functions content (are the actual functions clear enough, should they be adapted into smaller pieces?)
- [ ] From the upgrade mechanics perspective (9A/9B)

This is already great because a single python instruction can rebuild the whole app and the whole app is cleaned by dropping cascade the app schema

image

domi4484 (approved):
@ponceta from the code perspective it looks good to me.
For the other questions I am missing some postgres expertise to anwer yet.
In any case whatever comes out from the other questions, this changes doesn't put any limit in future decisions and should be safe to merge

datamodel/app/triggers/create_triggers.sql
```diff
@@ -0,0 +1,73 @@
+-- only update -> insert and delete are handled by reach trigger
```
3nids, line 1: I'm not a fan of separating the code which create the trigger from the ones which create the trigger functions.

3nids (commented):
Apart from the trigger file creation, this looks good to me.

I am not sure if I missed something important regarding this, but the main point is function renaming to clearly separate symbology from core netowrk business logic, is this correct?

datamodel/app/triggers/create_triggers.sql
ponceta: I've brought back the triggers but organized respectively the functions with their triggers.

ponceta:
@cymed if you can leave a comment on this I would very much appreciate! :)

cymed:
I agree with 3nids on keeping triggers and trigger functions together. I like separating the base functions from the trigger functions and the triggers. 
As discussed in #438, we should also allow intermediate tables (although I don't know why we need these tables, a materialized view should do the job)

Issue #489, opened 2024-11-05
Shortcuts in dict_od_tables and as prefix for attribute-names in views


urskaufmann:
**Describe the bug**
In two cases we have different shortcuts for tables in dict_od_tables and in the vw_tww-tables:
For channel is "CL" in dictionary, but "ch" in your views (was already in QGEP like this).
For biol_eco_assessment is "BO" in the dictionary, but "ba" in your views.

Suggestion: Change the english shortcuts in the dict_od_tables to CH and BA instead of CL and BO.

sjib:
Yes, I have to make the changes in the creation of the dictionary tables if we want to change that.

 

BA is the shortcut_en for Class Badestelle (in qgep), so we should not change that to stay with unique shortcuts per language.

CL -> CH for Channel would be possible. 

 

@ponceta How much effort is it to change the view?

 

I suggest a general principle (to be integrated in the Developers guides): Views should be developed based on the sys dictionaries on not the other way round else we loose control over used shortcuts

ponceta:
This should be no big deal to adapt views in consequence. Better do it before any official release is published.

sjib:
@urskaufmann @ponceta ok like this? If yes I will make a patch for qgep also

sjib:
- [ ] Add requirement to the Developers Guide to use official shortcut_en from Dictionary table in view creation

ponceta:
Were are these official shortcut defined (except in the code of dictionary tables)

sjib:
![tww_sys dictionary_od_table](https://github.com/user-attachments/assets/6f45d023-45ca-4b12-af74-f391dce03fa7)

PR #494, opened 2024-11-13
Adapt shortcut_en for channel from CL to CH

sjib:


sjib:
Should solve https://github.com/teksi/wastewater/issues/489

urskaufmann (approved):

PR #496, opened 2024-11-21
fix selection filters

cymed:
Closing #463


domi4484 (approved):
Could you add a test case?
I think we don't have any tests with selections at the moment and it should be fairly easy to add one

cymed:
Saw the comment only after merging - test will follow

Issue #495, opened 2024-11-21
Check datatype control_cable.symbolori


sjib:
Wrong definition of symbolori see https://github.com/qwat/qwat-data-model/pull/388#issuecomment-2490725104

Wrong also in tww

PR #497, opened 2024-11-21
correct cable_point.symbolori

sjib:


sjib:
Should solve https://github.com/teksi/wastewater/issues/495
Bug found in qwat - porting to tww

ponceta (approved):

PR #500, opened 2024-11-26
Redo #292

ponceta:
Supersed https://github.com/teksi/wastewater/pull/292

Is not yet working.

It looks like removing main node is actually completely destroying the wastewater structure.

image

- [ ] Fix https://github.com/teksi/wastewater/issues/250

### DIFF
--- a/datamodel/app/functions/symbology_functions.sql
+++ b/datamodel/app/functions/symbology_functions.sql
@@ -746,24 +746,25 @@ CREATE OR REPLACE FUNCTION tww_app.symbology_on_wastewater_node_change()
   RETURNS trigger AS
 $BODY$
 DECLARE
-  co_obj_id TEXT;
-  affected_sp RECORD;
+  wn_obj_id TEXT;
+  affected_ne RECORD;
 BEGIN
   CASE
     WHEN TG_OP = 'UPDATE' THEN
-      co_obj_id = OLD.obj_id;
+      wn_obj_id = OLD.obj_id;
     WHEN TG_OP = 'INSERT' THEN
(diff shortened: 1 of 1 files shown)


cymed:
How does `rel_wastewater_structure_main_wastewater_node` look in your database?

cymed:
I found the error:

https://github.com/teksi/wastewater/blob/79cbc31d5e7dc5fe93fc99b9f80dbfd7533a9dfe/datamodel/changelogs/0001/03_tww_db_dss.sql#L3076

We should use ON DELETE SET NULL

cymed:
I recall that I had a heritage db where we had a "ON DELETE CASCADE" for this relation. 
I just tried in 2024.0.3 where we keep the wastewater structure

 @ponceta did your wastewater structure have a cover? if not, it is possible that vw_tww_wastewater_strucrure simply had no geometry, as there is nor a cover nor a node

sjib:
> I recall that I had a heritage db where we had a "ON DELETE CASCADE" for this relation. I just tried in 2024.0.3 where we keep the wastewater structure
> 
@cymed Do we have to change something in the create scripts for DSS? If yes I will write a pull request for that and adapt my creation script

cymed:
No, in the current db it is corrct

sjib:
@ponceta @cymed Is this ready for 2024.0 or should it be integrated with 2024.1?

ponceta:
> @ponceta @cymed Is this ready for 2024.0 or should it be integrated with 2024.1?

This is not ready and is not mandatory for 2024.X

PR #502, opened 2024-11-29
Fix default values

cymed:
in tww_app.ft_set_default_values(), I passes an incorrect argument to tww_app.get_default_values(text), using table name instead of column name

### DIFF
--- a/datamodel/app/functions/organisation_functions.sql
+++ b/datamodel/app/functions/organisation_functions.sql
@@ -80,7 +80,7 @@ BEGIN
 	  OR  (c.table_schema ='tww_app'
 	  and t.table_type = 'VIEW'))
     LOOP
-        EXECUTE format($$ ALTER %4$s %1$I.%2$I ALTER COLUMN %3$I SET DEFAULT tww_app.get_default_values('%2$s') $$, sch,tbl,col,ttp);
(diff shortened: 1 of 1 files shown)


ponceta:
image

Postgres Default values are well interpreted but NULL for our 

```
    fk_dataowner character varying(16) COLLATE pg_catalog."default" DEFAULT tww_app.get_default_values('wastewater_structure'::text),
    fk_provider character varying(16) COLLATE pg_catalog."default" DEFAULT tww_app.get_default_values('wastewater_structure'::text),
```

ponceta:
@cymed Souldn't it be 

```
    fk_dataowner character varying(16) COLLATE pg_catalog."default" DEFAULT tww_app.get_default_values('fk_dataowner '::text),
    fk_provider character varying(16) COLLATE pg_catalog."default" DEFAULT tww_app.get_default_values('fk_provider '::text),
```

?

ponceta:
OK I figure out how it is supposed to work and it seems to work fine.

For now it looks like it is not possible to remove such a value once it is defined.

@urskaufmann We will then be able to remove all these fields from the QGIS project as they are fully managed by the database.

ponceta (approved):

PR #504, opened 2024-12-03
correct sqlalchemy version - 2.038 does not exist yet

sjib:


ponceta (approved):

ponceta:
FYI, SQLALCHEMY_MINIMAL_VERSION is now defined as 1.4 which is maybe wrong.

QGIS deploys sql_alchemy 2.0.29 with the osgeo4w bundle.

sjib:
1.4 was set because of certain functions used in the interlis import/export code.

@ponceta please discuss with opengis

PR #514, opened 2024-12-20
bo_ instead of ba_ for biological_assessment

ponceta:
Fix QGIS project
Necessary for translations


urskaufmann:
bo_ is now translated

PR #531, opened 2025-01-16
Add aggregation view for channel and georeferenced channel damages

cymed:
I found having a view vw_tww_channel practical, especially when planning maintenance events. The view aggregates all reach geometries linked to a channel. For now, the view is read-only

### DIFF
--- a/datamodel/app/create_app.py
+++ b/datamodel/app/create_app.py
@@ -12,6 +12,7 @@ except ImportError:
 from pirogue import MultipleInheritance, SimpleJoins, SingleInheritance
 from triggers.set_defaults_and_triggers import set_defaults_and_triggers
 from view.vw_tww_additional_ws import vw_tww_additional_ws
+from view.vw_tww_channel import vw_tww_channel
 from view.vw_tww_infiltration_installation import vw_tww_infiltration_installation
 from view.vw_tww_measurement_series import vw_tww_measurement_series
 from view.vw_tww_reach import vw_tww_reach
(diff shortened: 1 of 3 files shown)


ponceta:
Thank you for this new functionnality!

image

@cymed do we have a way to control the attributes order of the created view?

As it is an aggregation of compoundcurves should it be a multicompoundcurve geometry?

cymed:
> As it is an aggregation of compoundcurves should it be a multicompoundcurve geometry?

Yes. LineMerge aggregates to singlepart if the aggregated lines touch, but in a database this is not always the case - better to use multipart

ponceta:
Just as a small context comment, the LineMerge function aggregates to multilinestring and generalizes curved elements :

image

But as long as it is used in read only it should not break anything.

ponceta (approved):

sjib:
New feature that could help in this :

https://www.qgis.org/project/visual-changelogs/visualchangelog340/#feature-add-linear-referencing-symbol-layer-type

cymed:
> New feature that could help in this :
> 
> https://www.qgis.org/project/visual-changelogs/visualchangelog340/#feature-add-linear-referencing-symbol-layer-type

The feature does allow using a fixed interval, a z or an M value, or all vertices. None of these options work for channel damages. The feature is mainly good for hydrology

cymed:
@ponceta I had changes after approval. Ready for review

ponceta (approved):

PR #539, opened 2025-01-22
add label and symbology custiomisation guide to docs

cymed:


urskaufmann:
Labeling is ok
Symbology: I'm not sure. Manhole symbology depends on the connected channels. First prio:  outlet-channel (outlet of the main node). If there are 2/more outlets (which is really an exception), then it's the PAA-channel, that matters. If there are 2/more outlet-PAA-Channels, then it's the usage_current (tww_symbology_order in tww_vl.channel_usage_current).
If there is no outlet, then it depends on the inlets and here again the same rules: first PAA before SAA, then usage_current with the same prio as in the outlet case. I do not see the case when to use a separate inflow_prio.

cymed:
> Labeling is ok Symbology: I'm not sure. Manhole symbology depends on the connected channels. First prio: outlet-channel (outlet of the main node). If there are 2/more outlets (which is really an exception), then it's the PAA-channel, that matters. If there are 2/more outlet-PAA-Channels, then it's the usage_current (tww_symbology_order in tww_vl.channel_usage_current). If there is no outlet, then it depends on the inlets and here again the same rules: first PAA before SAA, then usage_current with the same prio as in the outlet case. I do not see the case when to use a separate inflow_prio.

Symbology is exactly what is currently implemented. We prioritise inflowing channels with discharged wastewater over outflows so we can symbolise discharge points into culverted water bodies with the _usage_current "discharged_wastewater". If there is no incoming channel with discharged wastewater, we prioritise by outgoing over incoming and function_hierarchic over usage_current

urskaufmann (approved):

sjib (approved):

PR #540, opened 2025-01-29
Create pg2ili_abwasser per default

cymed:
Currently, we re-create the schema pg2ili_abwasser every time someone imports or exports data (unless the datamodel is already present - in this case we just truncate all tables). This means that:

1. We need to grant CREATE on the database to tww_user, which gives them access to a wide variety of potential misuses
2. We effectively limit Interlis Import/export to one role, as only the owner of pg2ili_abwasser (which is the user that made the first import/export) can create/truncate the corresponding tables

This PR creates the schema pg2ili_abwasser on db creation and grants rights to tww_user, limiting the damage a tww_user can make and allowing multiple users to use the schema


domi4484 (commented):
@cymed you should also change the parameter `recreate_schema` to False otherwise the TRUNCATE/CREATE will still be executed:

https://github.com/teksi/wastewater/blob/main/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py#L88

https://github.com/teksi/wastewater/blob/main/plugin/teksi_wastewater/interlis/interlis_importer_exporter.py#L180

ponceta (approved):

PR #544, opened 2025-02-11
Prevent view crashing if `ST_LENGTH(re.progression3d_geometry)` is null or 0

rbovard:
## General
 - [x] Fix a bug

## Describe your changes
Prevent view `vw_tww_reach` from crashing if ST_LENGTH(re.progression3d_geometry) is null or 0

- [x] CI Tests are green

## Checklist before merge
- [x] A review has been performed
- [x] Comments are resolved
- [x] Documentation is ready

### DIFF
--- a/datamodel/app/view/vw_tww_reach.py
+++ b/datamodel/app/view/vw_tww_reach.py
@@ -47,7 +47,7 @@ def vw_tww_reach(pg_service: str = None, extra_definition: dict = None):
           ELSE clear_height
         END AS width,
         CASE
-          WHEN rp_from.level > 0 AND rp_to.level > 0 THEN round((rp_from.level - rp_to.level)/ST_LENGTH(re.progression3d_geometry)::numeric*1000,1)
+          WHEN rp_from.level > 0 AND rp_to.level > 0 AND ST_LENGTH(re.progression3d_geometry) IS NOT NULL AND ST_LENGTH(re.progression3d_geometry) > 0 THEN round((rp_from.level - rp_to.level)/ST_LENGTH(re.progression3d_geometry)::numeric*1000,1)
           ELSE NULL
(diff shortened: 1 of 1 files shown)


cymed:
In what situation do you need to digitise a reach with a length of 0 m?

sjib:
Thank you for this improvement.

I think it is good if this case of length = 0 m is treated and does not create a crash of the plugin (mistakes can happen or we import data from another system and then have such cases - I have seen a lot of cases that you never thought about in the last 20 years). 
@rbovard Maybe a warning could be added for length = 0 - this would help to find such cases easily

rbovard:
That's it, if you import data from external sources it makes the view crash.

@cymed The error is division by zero, so we need to check if the value is >0 (not only not null with COALESCE no?)

@sjib What do yo mean about this warning? You want me do something in this PR?

sjib:
@rbovard This adaption of the view is maybe not the best place to add a warning. But we have similar issues with the length profile. So thinking where to add a check and what especially it should highlight on this might be a good thing.  Can you create an extra issue?

rbovard:
[Done](https://github.com/teksi/wastewater/issues/546).

cymed:
> That's it, if you import data from external sources it makes the view crash.
> 
> @cymed The error is division by zero, so we need to check if the value is >0 (not only not null with COALESCE no?)

True. Nullif would be smarter

ponceta:
@rbovard you can prevent it with a Not Null constraint on the reach table, but this is still a good thing to prevent the view from crashing.

rbovard:
@cymed Thanks for the hint, I've adapted the code.

@ponceta Ok, please open another issue if you think it's relevant.

ponceta (approved):

Issue #529, opened 2025-01-14
TWW Project: Aufgabe gescheitert : Lade tww/Documents...


urskaufmann:
**Describe the bug**
in teksi_wastwater.qgs is the following line 4x:            <editforminitfilepath>C:/Users/tww/Documents/GitHub/TEKSI_wastewater/project</editforminitfilepath>

when open the project, I get always:
2025-01-14T15:02:23     INFO    Aufgabe gescheitert : Lade tww/Documents/GitHub/TEKSI_wastewater/project
2025-01-14T15:02:23     WARNING    Fehler Das Protokoll "" ist unbekannt beim Laden von tww/Documents/GitHub/TEKSI_wastewater/project über HTTP
or in english qgis
2025-01-14T13:44:04     INFO    Task failed : Fetching tww/Documents/GitHub/TEKSI_wastewater/project
2025-01-14T13:44:04     WARNING    HTTP fetch tww/Documents/GitHub/TEKSI_wastewater/project failed with error Protocol "" is unknown

**Expected behavior**
no such Infos and warnings

**Desktop (please complete the following information):**
 - TWW version 2024.0.5
 - QGIS Version 3.34.12
 - OS Win10

**Additional context**

sjib:
After fix this does not appear on my system - should be ok.

```
2025-06-13T14:41:56     INFO    Projektdatei ist älter : Dieses Projekt wurde mit der QGIS-Version 3.34.12-Prizren gespeichert. Erneutes Speichern der Projektdatei bringt die Datei auf den Stand der Version 3.40.6-Bratislava und verursacht u.U., dass es mit älteren Versionen von QGIS nicht mehr verwendet werden kann.
2025-06-13T14:42:09     INFO    Task abgeschlossen : Lade "TEKSI Wastewater and General Wastewater Management Plan (GEP)"
```

Issue #525, opened 2025-01-10
TWW-Project: Error on layer vw_change_points


urskaufmann:
**Describe the bug**
Layer vw_change_points is not available. But I can load the layer when disabling the option "CH PK unicity" while adding the layer to the project.

The problem seems to be an error in data. I had three reaches with to and from reach_point.fk_wastewater_networkelement always to the same wastewater_node. Therefore I got this wastewater_node three times as change_point.

Anyway: an error in data should not make a layer unavailable...

**Expected behavior**
no error when opening a project

**Desktop (please complete the following information):**
 - TWW version 2024.0.5
 - QGIS Version 3.34.12
 - OS Win 10

**Additional context**
to be changed in the tww-project (or is there someting to change in the vw?)

sjib:
Other check: Zero lenght or reach see https://github.com/teksi/wastewater/issues/546

cymed:
Solved for me in 2026.1.0

PR #518, opened 2024-12-23
Some Fixes for the QGS project

ponceta:
## General
 - [x] Fix a bug
 - [ ] Add a feature
 - [ ] Maintenance / sustainability
 - [ ] Add Documentation

## Describe your changes

- [x] Project : Remove duplicated fk_dataowner fields

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] CI Tests are green
- [x] The documentation is up to date with the proposed change.
- [x] My work is ready for review

## Checklist before merge
- [ ] A review has been performed
- [ ] Comments are resolved
- [ ] Documentation is ready

PR #559, opened 2025-03-10
Update qgs project from @urskaufmann

ponceta:
9.3.2025

- since the default dataowner and provider works, it makes no sense to have the fk_datawoner and fk_provider as "Enforce not null constraint" (you have to enter manually, otherwise you can not press "OK")
Changed in vw_tww_wastewater_structure.wn_fk_dataowner
Changed in vw_tww_wastewater_structure.wn_fk_provider
Changed in vw_tww_reach.fk_dataowner
Changed in vw_tww_reach.fk_provider
Changed in vw_wastewater_structure.fk_dataowner / .fk_provider
Changed in catchment_area.fk_dataowner / .fk_provider

-in vw_tww_reach is 3x the field fk_dataowner and fk_provider in the attribut form. 2x deleted

- Error in vw_tww_reach: ws_fk_dataowner and ws_fk_provider is still missing

- because you can not export with wastewater_structure.fk_owner / fk_operator = NULL, it makes sense to set the "Not null" constraint (or even the "Enforce not null constraint").
Changed in vw_tww_wastewater_structure.fk_owner
Changed in vw_tww_wastewater_structure.fk_operator
Changed in vw_tww_reach.fk_owner
Changed in vw_tww_reach.fk_operator
Changed in vw_wastewater_structure.fk_owner / .fk_provider

- in relation reference, the Allow NULL value has to be checked, because otherwise QGIS shows a value in the attribute form, also if there is no entry in this field
Changed in vw_cover.fk_wastewater_structure
Changed in vw_wastewater_node.fk_wastewater_structure

- vw_change_point inserted in project without CH PK unity (#525)

- cover warnings placed next to the cover attributes.

- It would make sense, to pre-define the organisation "unbekannt", obj_id ch20p3q400000107. So the user has always an option even if he does not know the owner.

13.06.2025

- [x] Fix https://github.com/teksi/wastewater/issues/529
- [x] Supersed https://github.com/teksi/wastewater/pull/518


cymed:
@ponceta Can you merge the changes into #518? We should only have one open PR altering the qgs project at a time to not run into problems

sjib:
- [ ] Could this be fixed also before merge? https://github.com/teksi/wastewater/issues/529

in teksi_wastwater.qgs is the following line 4x: C:/Users/tww/Documents/GitHub/TEKSI_wastewater/project

Issue #558, opened 2025-03-09
Drainage chanel too wide


urskaufmann:
**Describe the bug**
When drawing a drainage chanel, the width of the polygon is fix coded (0.4m). Normal drainage chanels have a width of 10 to 20cm (but not 40cm).
The value is defined in plugins\teksi_wastewater\tools\twwmaptooladdfeature.py (last 40 lines, 3 x 0.2). Please change the value to 0.1 (so the drawing will have a width of 0.2m).

To top solution would be: when drawing the line to define the polygon, I can press option-key and can enter the width of the polygon (default = 0.2).

ponceta:
![Image](https://github.com/user-attachments/assets/167cbb88-0ebb-4144-8565-713e373fd296) 

You can change the width with the CTRL clic option before click on the second point but it is an easy fix to change it to 0.2m per default.

![Image](https://github.com/user-attachments/assets/dbe5b65b-ab85-4b1d-8b41-7c9ddeea64cb)

The second blue message does not appear so there's still a bug to find!

urskaufmann:
Thanks and sorry, its even described in the docu, I didn't remember. 
Not correct is: you have to enter the half width! If you enter 0.1, then the polygon has a width of 0.2.
So my default would be: a polygon-width of 0.2m, means 0.1 in the twwmaptooladdfeature.py. and a user entry that has to be halved  for the point calculation

sjib:
> The second blue message does not appear so there's still a bug to find!

@urskaufmann  What is the second blue message? the one you copied above?

https://github.com/teksi/wastewater/blob/82ad6632ceafebea13a23dde6b70ebf83595b072/plugin/teksi_wastewater/tools/twwmaptooladdfeature.py#L440

urskaufmann:
The second blue message does not appear so there's still a bug to find!

When digitizing drainage channel, the blue message is ok. The blue message is still missing when chosing action Digitize

sjib:
may be this could be added in the qgs file, as with "Digitize" not an extra TWW is called, but a standard QGIS Function:

  ```
[...]

PR #561, opened 2025-03-12
2025-03-25-fix-default-digitize-drainage-channel

sjib:
change default to 01. instead of 0.2, divide user entry by 2


sjib:
Should solve https://github.com/teksi/wastewater/issues/558

sjib:
@urskaufmann Can you check if this works now as expected?

urskaufmann (changes requested):
txt = QLineEdit("0.2") 

then it's ok.

ponceta (approved):

Issue #566, opened 2025-03-23
catchment area not selected with "Select current path"


urskaufmann:
**Describe the bug**
one of the missing feature is, that the user has no tool to select the catchment areas, that where connected to selected nodes. That's why I search for code in the current plugin. And I found in \python\plugins\teksi_wastewater\gui\twwprofiledockwidget.py  some code, that looks like tww should select not only nodes, but also catchment areas when choosing the Select current path command. Line 173ff

But in line 173 is code that can not work, because it is from qgep. 
catchment_areas_layer = TwwLayerManager.layer("od_catchment_area")
In qgep we had layers with od_*. !

It works, if I change this line to 
catchment_areas_layer = TwwLayerManager.layer("catchment_area")

**To Reproduce**
Exact steps to reproduce the behavior:
- use the upstream network following tool
- click on "Select current path"
- no catchment area are selected

**Desktop (please complete the following information):**
 - TWW version 2024.0.5
 - QGIS Version 3.40.4
 - OS Win 10

Little change to use the code, that already exists and to help me on my daily work.

The help would be bigger, if the user can select the nodes and catchment_areas starting with a selection of reaches (separat tool)

cymed:
Thanks for pointing that out! Could you provide a pull request?

ponceta:
For the future, we should ensure that layer names syntax or translations are not affecting actions or plugin.

PR #567, opened 2025-03-25
Update twwprofiledockwidget.py

urskaufmann:
solves #566


sjib:
@urskaufmann looks good to me. Can you request my review?

sjib:
I guess this will only work in the english project file - as long as we do not add some code for translated layer names. Does anyone knows if QGIS made any progress on this?

See https://github.com/QGEP/QGEP/issues/839

urskaufmann:
I did test this with german and english project file. It works on both project (the selection of reaches and node do also work on a translated project).

ponceta (approved):

ponceta:
Thank you @urskaufmann for this catch!

Issue #576, opened 2025-03-31
Images broken in documentation at teksi.github.io


w0pr:
Dangling image URLs break documentation:

![Image](https://github.com/user-attachments/assets/79941068-bea3-4159-98bb-285a471d8975)

![Image](https://github.com/user-attachments/assets/d75caf11-2217-4325-bb3c-b6d820f1b9e2)

ponceta:
https://teksi.github.io/wastewater/user-guide/layerexplanations/layerexplanation.html This looks to be fixed by @sjib, thank you @w0pr for your feedback!

w0pr:
> https://teksi.github.io/wastewater/user-guide/layerexplanations/layerexplanation.html This looks to be fixed by [@sjib](https://github.com/sjib), thank you [@w0pr](https://github.com/w0pr) for your feedback!

Still at least one left:
https://teksi.github.io/wastewater/admin-guide/db-production-readiness/index.html#productionreadiness

![Image](https://github.com/user-attachments/assets/d04f2501-54d1-43b9-be75-9eb743aeb146)

sjib:
> Still at least one left: https://teksi.github.io/wastewater/admin-guide/db-production-readiness/index.html#productionreadiness
> 
> ![Image](https://github.com/user-attachments/assets/d04f2501-54d1-43b9-be75-9eb743aeb146)

This should fix it: https://github.com/teksi/wastewater/pull/583

PR #577, opened 2025-03-31
2025-03-31-fix-images-docs

sjib:
rename images to fit tww


sjib:
Should solve https://github.com/teksi/wastewater/issues/576

ponceta (approved):

PR #595, opened 2025-05-01
Fix type 'unknown' handling for vw_tww_wastewater_structure

cymed:
When updating a wastewater structure with no subclass or a subclass that is not handled by vw_tww_wastewater_structure, the update fails due to lacking `ELSE NULL;` clauses. This PR fixes that


ponceta:
@cymed could you please provide the same fix for 

https://github.com/teksi/wastewater/blob/main/datamodel/app/view/vw_tww_additional_ws.py?

ponceta (changes requested):
See previous comment, otherwise looks good to me.

ponceta (approved):

ponceta:
@cymed thank you!

PR #598, opened 2025-05-06
Add extension framework

cymed:
I already redeveloped the extension framework in #399, but realised it is smarter for further development if the framework can be merged separately from the AG-xx extension (CI compatibility). 

The main changes are the following:

1. `create_app.py` is now a module
2. Instead of passing all extra_definition dicts, these dicts can be passed as extensions.
3. The argparse call of `create_app.py` allows passing a list of extension names (argument `extension_names`) that are loaded.
4. The structure of the extension definition are stored in \datamodel\app\extensions\config.yaml
5. We now run all plain sql files in certain folders. This means we don't need to adapt `create_app.py` when adding a script

Furthermore, I added the following tests to CI:

-  Test to check whether it is possible to initialise an extension (psycopg 2 and 3)
-  Test to check whether any extension creates a table in tww_od
-  Prepared test to run unit tests for individual extensions (once we start merging them)


ponceta:
The changes in this diff reflect a significant restructuring and enhancement of the `create_app` functionality. Here's my assessment:

### The Good:
1. **Improved Modularity**:
   - Introduction of `load_extension` and a robust `extension_names` parameter make the code more modular and extensible.
   - The use of `yaml_data_dicts` centralizes YAML-based configurations, reducing redundancy.

2. **Simplified SQL Execution**:
   - Replacing multiple `run_sql_file` calls with `run_sql_files_in_folder` simplifies the execution of SQL scripts.

3. **Better YAML Handling**:
   - The `load_yaml` function safely loads YAML files with error handling, ensuring robustness.

4. **Scalable Structure**:
   - New features like `MultipleInheritances`, `SingleInheritances`, and `SimpleJoins_yaml` make the code easier to extend for new views.

5. **Cleaner CLI Interface**:
   - Removing individual YAML file CLI arguments and replacing them with the `extension_names` parameter simplifies the command-line interface.

### Concerns:
1. **Verbose Logs**:
   - The extensive use of `print` statements for debugging and logging (e.g., `Running extension {extension}`) might clutter logs in production. A proper logging framework would be more appropriate.

2. **Backward Compatibility**:
   - The removal of CLI arguments like `--tww_reach_extra` and `--tww_wastewater_structure_extra` may break existing workflows if not properly communicated to users.

3. **Potential Overhead**:
   - Additional complexity with features like `MultipleInheritances` and `SimpleJoins_yaml` may confuse new developers or users who are not familiar with the codebase.

4. **Testing Extensions**:
   - The introduction of `demo_extension` files like `001_demo.sql` and `002_demo.py` is good for demonstration, but it's unclear if there's a robust framework for testing these extensions.

### Areas for Improvement:
- **Logging Framework**: Replace `print` statements with a proper logging mechanism (`logging` module in Python).
- **Documentation**: Ensure the new structure and CLI changes are well-documented for both developers and end-users.
- **Error Handling**: Ensure that failures in loading extensions or YAML files are gracefully handled and do not break the entire application.

### Overall:
It looks like a well-thought-out update, focusing on scalability and modularity. With a few tweaks for logging, error handling, and backward compatibility, it can be considered quite "pretty"!

ponceta (approved):
New functionnality with a long awaited structure to handle extensions properly.

++ Will still evolve in the future but a great start for our needs.
-+ Requires some coding skills.

image

3nids:
I had a first look. 
Just to be sure: this is limited to app changes only, is this correct?

The idea of having a definition of extensions and having this list passed as a parameter sounds good.
But if extensions have parameters too, we have to define a way to pass them from the beginning (from create_app).

More on the code:
* why do we need an argument parser in the extension since we should not be running from the command line?
* it seems there is an overwriting process of the views which is not illustrated in the demo. Is the demo planned to stay in the code? if yes, it would be great to have a demo for all the functionality and have it tested.

I'm a bit concerned about adding an extension framework just for views and bundled in the TWW app. Ideally we would have a standard way to define extensions that have both data and app additions and that would be handled from pum. 

Just to be sure, what are the requirements for these extensions: add new views and replace existing views?

One question that is really important to me at the moment is the removal of all the extra definitions parameters. I don't recall when and why they were added. Does anyone recall? Can they safely be removed? (that would help)

sjib:
@cymed 

Adding custom triggers
""""""""""""""""""""

```
It is possible to create custom views and add them to the data model.
Simply create an extension folder, place the sql or py script inside and add the ``--extensions`` flag
Please note that these scripts are re-run on every datamodel update.

```
I think this needs adaption for triggers - it is copy paste from views

cymed:
Waiting for https://github.com/opengisch/pum/issues/122 to be resolved

datamodel/changelogs/2025.0.2/02_extensions_agxx_od.py
```diff
@@ -0,0 +1,83 @@
+try:
+    import psycopg
+except ImportError:
+    import psycopg2 as psycopg
+
+from pum.core.deltapy import DeltaPy
```
3nids, line 6: In pum 1.x there is no more python deltas, they need to be pure SQL (with parameters)

PR #602, opened 2025-05-14
Minor fix 12_1_roles.sql

cymed:
The script [12_1_roles.sql](https://github.com/teksi/wastewater/blob/main/datamodel/12_1_roles.sql) does not work when there is a dash in the database name. This PR fixes that.


ponceta (approved):

ponceta:
Why removing grants on tww_cfg?

cymed:
Because we are removing tww_cfg anyway

Issue #601, opened 2025-05-13
Reach_point without geometry: error or not?


urskaufmann:
**Describe the bug**
Import of DSS 2020 data without error (ilivalidator is valid). 
But there are some reach_points without geometry (just with level).
This fact gives then a bit later a orange message Python-error:

_WARNING    Traceback (most recent call last):
              File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_de/python/plugins\teksi_wastewater\tools\twwnetwork.py", line 112, in _addVertices
              vertex = feat.geometry().asPoint()
              ^^^^^^^^^^^^^^^^^^^^^^^^^
             ValueError: Null geometry cannot be converted to a point.
             
             During handling of the above exception, another exception occurred:
             
             Traceback (most recent call last):
              File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_de/python/plugins\teksi_wastewater\teksi_wastewater_plugin.py", line 495, in refreshNetworkTopologyActionClicked
              self.network_analyzer.refresh()
              File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_de/python/plugins\teksi_wastewater\tools\twwnetwork.py", line 194, in refresh
              self.createGraph()
              File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_de/python/plugins\teksi_wastewater\tools\twwnetwork.py", line 219, in createGraph
              self._addVertices()
              File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_de/python/plugins\teksi_wastewater\tools\twwnetwork.py", line 114, in _addVertices
              raise ValueError(f"No Point Geometry found for Node {obj_id} (Type: {obj_type})")
             ValueError: No Point Geometry found for Node ch19jxdjgsplNnyQ (Type: reach_point)_

Trying to refresh network topology (SQL-Button): 

 _WARNING    Fehlerhafte Abfrage: SELECT tww_app.network_refresh_network_simple(); ergab 7 [ERROR: canceling statement due to statement timeout
             CONTEXT: SQL statement "REFRESH MATERIALIZED VIEW tww_app.vw_network_node"
             PL/pgSQL function tww_app.network_refresh_network_simple() line 106 at SQL statement_

After all, Export to DSS is possible without error (valid).

**Expected behavior**
[...]

PR #614, opened 2025-06-03
Add substitute for reach point geometry in network node

cymed:
As described in #601, VSA DSS allows reach points with NULL geometry. These break the network functions because we need a geometry for the network tree. This PR uses the start/end node of the reach that references the reach point as a fallback value when generating the tree.

### DIFF
--- a/datamodel/app/functions/network_functions.sql
+++ b/datamodel/app/functions/network_functions.sql
@@ -20,7 +20,7 @@ BEGIN
     rp.obj_id,
     ST_Force2D(coalesce(rp.situation3d_geometry,ST_EndPoint(r_t.progression3d_geometry),ST_StartPoint(r_f.progression3d_geometry)))
   FROM tww_od.reach_point rp
-  LEFT JOIN tww_od.reach r_f ON rp.obj_id = r_f.fk_reach_point_from 
+  LEFT JOIN tww_od.reach r_f ON rp.obj_id = r_f.fk_reach_point_from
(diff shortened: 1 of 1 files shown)


cymed:
This probably means the query took too long to load. 
I just saw that we have no indexes on `tww_od.network_node` and `tww_od.network_segment`. Could be part of the problem.

w0pr:
> Test: Existing database, all reachpoints with geometry. I alter the function in pgadmin. I import then another database with missing geometry on some reachpoints with TWW-import tool. Import works, no error. TWW-Project no error. TWW-SQL-Button no error. Network-following works. The new function works. The timeout-problem does exists, if there is missing geometry and old function in the database (in the tww_app.vw_network_node )

I don't believe the timeout error is a result of missing geometry, this sound like a database client (pgadmin) issue; there should be a way to configure client timeout (set it to infinite). Rebuilding materialized views can take *hours* in a reasonably large database.
I use dbeaver and never ran into a timeout (but watched an active session with the query "select tww_app.network_refresh_network_simple();" chugging away for hours).

Also, building those materialized views with NULL geometry works just fine, it is the dependent network fuctions that will complain.
This PR will fix the geometry issue, but rebuilding the views will arguably take even a little bit longer now with the extra code (joins, etc.)

urskaufmann (approved):
As I wrote, its works. What I saw, is that I had to run this function first manually, if there have been null-geometry-reachpoints in the database.

Issue #586, opened 2025-04-09
Make SRID configurable throughout project


w0pr:
Although documentation states that [the SRID can be set during database initialization](https://teksi.github.io/wastewater/installation-guide/database-initialization.html#generate-the-data-model-under-linux), the "-s" parameter (or any parameter) is not even evaluated in setup.sh and in various places the default swiss SRID 2056 is hardcoded (e.g. plugin code and ili2pg parameters, geometry_functions.sql, etc.), making it hard to use a different coordinate system:

https://github.com/search?q=repo%3Ateksi%2Fwastewater+2056+&type=code

![Image](https://github.com/user-attachments/assets/eab441e1-a231-4921-9374-036ba89e9368)

Also name of script isn't db_setup.sh, but setup.sh.

w0pr:
> [@w0pr](https://github.com/w0pr) Happy if you make a pull request to adapt this as suggested.

I'll give it a try, DB scripts are straightforward; plugin code seems more involved.

ponceta:
The more generic the better, start with small propositions so they can be merged in an easy way.

w0pr:
> The more generic the better, start with small propositions so they can be merged in an easy way.

Of course. Then again, the QGIS project (all datasources, etc. are also fixed with SRID 2056), there is no way I know of how to change this for all (other than manually, which I did for my project), but there's no way to parametrize this.

ponceta:
QGIS project parametrization is a nightmare right now, I come back to you if we find any hint on this with the installation plugin we're currently developing.

ponceta:
Even for the datamodel we used dirty hacks to provide some SRID customization on the old wastewater module (QGEP) : https://github.com/QGEP/qgepplugin/blob/master/qgepplugin/gui/qgepdatamodeldialog.py#L724

cymed:
Please note that in #399, the SRID is passed as a raw variable for all plain sql scripts.

w0pr:
> Please note that in [#399](https://github.com/teksi/wastewater/pull/399), the SRID is passed as a raw variable for all plain sql scripts.

![Image](https://github.com/user-attachments/assets/ed7b6bbe-6681-4961-8bc0-3015700be570)

Ah, ok. In current release, there are still some hardcoded instances of SRID, I'll look into this PR, because I have to think about maybe creating extension for german communal wastewater anyway. Thanks.

cymed:
[...]

PR #563, opened 2025-03-14
Revert #538  "fix tww_symbology_inflow_prio"

cymed:
Reverts teksi/wastewater#538

PR #617, opened 2025-06-05
Check for alterations in already released datamodel versions

cymed:
This PR adds two tests 

 1. checks if there were any changes in datamodel/changelogs/2025.0.1. We want to forbid alterations there as they can corrupt the db update with pum.
 2. checks for references to tww_app in the pum update scripts.


cymed:
#563 needs to be merged before this PR

sjib:
Helps solve https://github.com/teksi/wastewater/issues/586

.github/workflows/check-changes-in-changelog.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Check for changes in the specified folders
+        run: |
+          LATEST_STABLE_RELEASE="2025.0.1"
```
3nids, line 16: This should be done programmatically using gh CLI.

.github/workflows/check-changes-in-changelog.yml
```diff
@@ -0,0 +1,27 @@
+name: Check for changes in stable datamodel versions
+
+on: [push, pull_request]
```
3nids, line 3: I would restrict the branches, otherwise this job will be run twice on PRs.

.github/workflows/check-changes-in-changelog.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Check for changes in the specified folders
+        run: |
+          LATEST_STABLE_RELEASE="2025.0.1"
+          CHANGED_FILES=$(git diff --name-only HEAD^ HEAD)
```
3nids, line 17: This will result in checking only the last commit, so this won't work as is. Easy solution is to use  tj-actions/changed-files (which has been compromised in march, but can still be recommended)

3nids (commented):
General idea is good but it needs a few adaptions.

.github/workflows/check-changes-in-changelog.yml
cymed: Is there a specific version you'd suggest

3nids: taking the last one sounds good v46.0.5

3nids (approved):
nice! thanks for following up!

PR #618, opened 2025-06-05
replace hardcoded 2056 SRID with :SRID parameter in changelogs/0001 SQL scripts

w0pr:


sjib:
Helps solve https://github.com/teksi/wastewater/issues/586

sjib:
I think there are some more:
grafik

w0pr:
> I think there are some more: grafik

right, but not inside changelogs/0001

ponceta (approved):
LGTM, thank you for providing this!

PR #622, opened 2025-06-10
update ilivalidator to 1.14.7

ponceta:
update ilivalidator to 1.14.7


ponceta:
@domi4484 Iooks like pum evolution is breaking CI, I wait for its evolutions before going further.

PR #509, opened 2024-12-17
Separate symbology and labels from od

cymed:
## General
 - [ ] Fix a bug
 - [x] Add a feature
 - [x] Maintenance / sustainability
 - [x] Add Documentation

## Describe your changes
Concept of #409, but including symbology. 

The PR introduces 
- a table `tww_od.tww_reach_point_labels`
- a table `tww_od.tww_wastewater_structure_labels`
- a table `tww_od.tww_wastewater_node_symbology`
- a view `tww_app.vw_tww_wastewater_node` to include the now separated symbology columns. 
- TODO: currently, moving the `vw_tww_wastewater_node` moves the corresponding reaches/reach points as well. Is this behaviour desired?

## To-Do
When altering a geometry in `tww_app.vw_tww_wastewater_node`, the geometry of the reaches/reach points linked to it are updated too. We need feedback from the user group if this behaviour is desired or not.

Also, the QGIS project needs to be updated (labels, wastewater node view,...)

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] CI Tests are green or the error can be ignored
- [x] The documentation is up to date with the proposed change.
- [x] My work is ready for review

## Checklist before merge
- [ ] A review has been performed
- [ ] Comments are resolved
- [ ] Documentation is ready

PR #625, opened 2025-06-13
Pum integration

domi4484:


3nids (approved):

sjib:
In what format would you need demodata?

3nids:
> In what format would you need demodata?

I don't need demodata :)

cymed:
Some additional questions:

1. Why was the auditing tool disabled? see #249
2. Why were the changes to the app schema in #598 not taken into consideration before implementing this PR? It re-alters the whole structure of tww_app, leading to unnecessary work for #598 and #399
3. Where is the documentation on how to define datamodel updates?We have pending od changes in #120, #292, #399, #509, #537, #540, #563, #612, #615

3nids:
The changes here are indeed required by the integration of pum/oqtopus/TMMT.
Mainly:
- the folder of the changelogs has been renamed to the version (`0001` => `2025.01`)
- no transaction commits in changelogs or app are allowed (to allow rollback in case of failure during upgrade)
- create_app now inherit [pum.HookBase](https://opengisch.github.io/pum/api/hook_base/)
- variables (SRID) are now written in the Python form (`{SIRD}` instead of `:SRID`)

While there is a lot of files modified there is no "big" change in the code and there is strictly no change in the datamodel (except of the addition of the `pum_migrations` table)

3nids:
@ponceta description added to the top message

@cymed 
> Why was the auditing tool disabled?
it will be reintroduced soon, it still requires a few adaptions (triggers in app schema)

> Why were the changes to the app schema in https://github.com/teksi/wastewater/pull/598 not taken into consideration before implementing this PR?
Work was done at the same time, and its integration would also have required a rebase on the pum integration work. I believe this still needs some discussion and it might be easier once the pum integration is achieved. In any case, I believe the rebase effort should not be too difficult (but it's always painful I totally agree)

> Where is the documentation on how to define datamodel updates?
There is the pum documentation for this: https://opengisch.github.io/pum/getting_started/#changelogs
It might be still a bit light, would be great to hear feedback
I think most of these PRs are more-or-less in a ready to be merged state (after a simple rebase for most of them)

ponceta:
@3nids Thank you!

PR #633, opened 2025-06-23
Extension framework

cymed:
Due to problems with the rebase, new PR instead of #598


datamodel/app/create_app.py
```diff
@@ -12,62 +16,84 @@
...
         """
         Creates the schema tww_app for TEKSI Wastewater & GEP
         :param SRID: the EPSG code for geometry columns
-        :param tww_reach_extra: YAML file path of the definition of additional columns for vw_tww_reach view
-        :param tww_wastewater_structure_extra: YAML file path of the definition of additional columns for vw_tww_wastewater_structure_extra view
-        :param tww_ii_extra: YAML file path of the definition of additional columns for vw_tww_infiltration_installation_extra view
-        :param wastewater_structure_extra: YAML file path of the definition of additional columns for vw_wastewater_structure_extra view
-        :param tww_channel_extra: YAML file path of the definition of additional columns for vw_tww_channel_extra view
+        :param extension_agxx: bool of whether to load agxx extension
+        :param extension_agxx: bool of whether to load ci extension
```
ponceta, line 39:
```suggestion
        :param extension_ci: bool of whether to load ci extension
```

datamodel/app/create_app.py
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env python3
-
+import copy
+import os
+import re
+import zipfile
```
ponceta, line 5: Do we have a way not to have all these imports if we don't play with extensions?

ponceta (commented):
Looks good to me, a big extension to the actual framework but could be a great plus if we have to manage complex extensions like ag64

datamodel/app/create_app.py
cymed: see 6e4459c 
os and re are always needed, as they are used to run all sql files in a specific folder and to parse variables

3nids (commented):
I made some minor remarks in the code. This is generally well coded.

I have some concerns with the definition of "extension". The current approach is built on the assumption that an extension is always defined like this: everything in app (no od additions), modification of views. While this works for AGxx, this might not be true for other needs.

One approach would be to drop the extension framework and simply do an `if agxx_enabled`. This would simplify the code but I would not recommend this since the work done here offers here a nice flexibility for testing: no need to touch the code base to test the modifications of views which I think is also really worth it. But also a good flexibility for "standard users" to alter the views without modifying the code. This is really a super neat addition.

So, my recommendation is to stick to what is proposed here but keeping in mind that it **might** block us in the future.

I would just recommend to have the extension definition more explicit and declarative. Each extension would define its own yaml config file. It would contains:
* the parameters
* the list of actions to do, probably defined as
  * pre/post (before or after the standard app creation)
  * within a choice of action (run_sql, overwrite view, create view, etc).

A remaining question is the correct integration with pum and oqtopus for parameters definition. But I think it is easy to just replicate the core extension parameters in the pum config file.

.github/workflows/datamodel-test.yml
```diff
@@ -56,6 +56,71 @@ jobs:
...
+  #       run: |
+  #         docker run -d -p 5432:5432 --name teksi-wastewater -e EXTENSION_NAMES=${{ matrix.extension }} teksi/wastewater:unstable
+  #         docker exec teksi-wastewater init_db.sh wait
+
+  #     - name: Run tests
+  #       run:  docker exec teksi-wastewater pytest datamodel
+
+  #     - name: Docker logs
+  #       if: failure()
+  #       run: docker logs teksi-wastewater
```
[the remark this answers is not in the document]
cymed: I will activate the test once I merge agxx

datamodel/app/create_app.py
```diff
@@ -114,119 +140,278 @@ def run_hook(
...
+    def parse_variables(self, variables: dict) -> dict:
+        """Parse variables based on their defined types in the YAML."""
+        formatted_vars = {}
+
+        for key, meta in variables.items():
+            if isinstance(meta, dict) and "value" in meta and "type" in meta:
+                value, var_type = meta["value"], meta["type"].lower()
+
+                if var_type == "raw":  # Directly insert SQL without escaping
+                    formatted_vars[key] = psycopg.sql.SQL(value)
```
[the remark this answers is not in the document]
cymed: Just tried as literal, it breaks the code. Needed to pass ints. I added a regex to only pass numeric values

3nids:
> there are vl alterations, too. The od extensions of official extensions are loaded for everyone

OK, so I believe there is something wrong in the design since altering the data schemas is not allowed at the app creation phase. In the AGxx PR, you create VL tables from the app. why?

> thus the option to pass a zip file containing the yaml

Even more, I would go to only point a YAML file (which then points itself to its needed files). That would remove the zip dependency and lower a bit the code.

> Separating the yaml and adding pre/post migration hooks per extension would be fine with me. It is however something we can do when tackling oqtopus integration.

It is really the core design of this extension framework and is independent from oqtopus. 

> The suggestion of the basic framework stood before the development start of oqtopus, so I'd rather optimise it there.

I would personally recommend to align as much as possible to avoid bringing something not finalized in the code base.
After reviewing, it sounds that having this code in pum rather than in tww would allow a much more robust and integrated experience. This can be done later for sure but it will be much more complicated.

In the end I see 3 options:
* merge as is
* make the changes here (yaml) to align a bit more with pum approach
* move the extension logic to pum

Moving to pum would probably solve the issues to correctly handle (data vs app phases, offer an integrated approach for parameters).

datamodel/app/app_modification.template.yaml
```diff
@@ -0,0 +1,42 @@
...
+  vw_tww_wastewater_structure: null
+  vw_tww_overflow: null
+  vw_wastewater_structure: null
+  vw_tww_infiltration_installation: null
+  vw_tww_channel: null
+  vw_tww_damage_channel: null 
+  vw_tww_additional_ws: null
+  vw_tww_measurement_series: null
+        
+multiple_inherintances:
```
sjib, line 16: Müsste `multiple_inheritance` heissen - Mehrfachvererbung

cymed: es sind mehrere Mehrfachvererbungen

sjib: Ja, dann multiple_inheritances statt multiple_inherintances

cymed:
> > Decision discussed by @ponceta @3nids @cymed:
> > ```
> > * The extension framework is renamed to modification framework
> > 
> > * the modification framework must not alter od tables
> >   
> >   * alterations to od are handled in the changelogs
> >   * extensions do not add attributes to existing VSA tables, but use 1:1 relations from their own tables instead
> > 
> > * the modification framework may activate/deactivate vl entries
> >   
> >   * if an extension needs additional vl, they are added in changelogs
> >   * extension vl entries are inactivated by default
> > 
> > * Instead of passing alteration yaml to create app, there will be a template yaml that defines how alterations are handled.
> > ```
> 
> Can these decisions be integrated in the https://github.com/teksi/Home/wiki/TEKSI-Developer-Guide and the TMMT Documentation [teksi/TMMT#21](https://github.com/teksi/TMMT/issues/21)

Yes, once the solution works

datamodel/app/view/utils/extra_definition_utils.py
```diff
@@ -0,0 +1,114 @@
...
+                )
+                for table_def in (
+                    t
+                    for t in extra_definition.get("joins", {}).values()
+                    if not t.get("read_only", True)
+                )
+            ]
+        )
+    except Exception as e:
+        print("An error occurred:", e)
```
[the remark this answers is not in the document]
cymed: removed the try - catch

datamodel/app/modification_repos/tww2ag6496/config.yaml
```diff
@@ -0,0 +1,26 @@
+variables:
+   lang_code:
+     value: de
+     type: literal
+   SRID:
+     value: 2056
+     type: number
+
+sql_files:
+  - file: sql/01_defaults.sql
```
ponceta, line 10: Are these files only in your repo / branch?

ponceta (approved):
Looks it is going in the direction of our discussions. Ok for me!

datamodel/app/modification_repos/tww2ag6496/config.yaml
cymed: They are added in #634

PR #639, opened 2025-06-24
allow base36 encoding for OID postfixes

cymed:
Tackling https://github.com/QGEP/QGEP/discussions/764 and https://github.com/teksi/wastewater/discussions/635
This PR introduces base36 encoding for OID postfixes >999999. There is a jump from 999999 to 559744029 to ensure compatibility with the old enumeration system. 

ch000000ws999999 is followed by 
ch000000ws99999a and then it follows base36 behaviour (as there are no digits-only numbers left anymore). This gives us roughly 1.6 billion valid postfixes per class.


sjib (approved):
Great improvement.
Where could we describe the new behaviour and document how generate_oid is working?
Would a new chapter in the Admin guide be helpful that explains the concept how obj_id's are generated?
https://teksi.github.io/wastewater/admin-guide/index.html
@w0pr @ponceta @cymed

PR #648, opened 2025-07-01
2025-07-01-typo_function_hierarchic

sjib:
SWMM Documentation typo_function_hierarchic


sjib:
@ponceta Why do we have this error on the CI?


```
Invalid workflow file: .github/workflows/docs-builder.yml#L27
The workflow is not valid. .github/workflows/docs-builder.yml (Line: 27, Col: 11): Input release_tag is required, but not provided while calling.
```

ponceta:
> @ponceta Why do we have this error on the CI?

CI issue, I fixed it yesterday

PR #649, opened 2025-07-01
2025 07 01 improve maintenance event views

sjib:
This improves the two views with 

- [x] additional attributes 
- [x] and a function `ft_vw_tww_maintenance_on_reach_wws_update`
- [x] renaming to vw_tww_maintenance_event_on_reach.sql and vw_tww_maintenance_event_on_wws.sql
- [x] adapt to official shortcuts for classes as defined in tww_sys.dictionary_od_table

as basis for easy visualisation and report printing of all three maintenance_event subclasses (maintenance (new in 2020), examination (for VSA-KEK TV Inspections) and bio_ecol_assessment (assessment at discharge_points into rivers)


urskaufmann:
Examples of work with these views (maintenance with geometry):

Visualize Unterhaltsetappen:
image

Visualize planned or realized renovations:
image

Select manholes/reaches based on a selection of maintenance events:
image

cymed:
`vw_tww_channel` was introduced in https://github.com/teksi/wastewater/releases/tag/2025.0.1rc2
We'd need to alter the view to accomodate linked maintenance events

cymed:
depends on #633

urskaufmann:
- I miss the fields with the identifier of the from_node and to_node.
- Is there a way to get the text of the material (in the list are just the code-numbers)

cymed:
The materials are listed as an array, and for now, there is no option to apply value relations to array types in QGIS

sjib:
@cymed Can you print a sample of the parsed view?

urskaufmann:
> > * I miss the fields with the identifier of the from_node and to_node.
> 
> Do you mean the vonPunktBezeichnung / bisPunktBezeichnung? They are in examination, not in maintenance_event

no, I mean vonSchacht - bisSchacht. I use this view to export the liste of maintenance events (notwendige Sanierungen). In this list is also the material important, because there are repair methods that do not fit with all material.

datamodel/app/view/maintenance_views.py
```diff
@@ -177,7 +177,7 @@ def vw_tww_channel_maintenance(connection: psycopg.Connection, extra_definition:
         , {ch_cols}
         {extra_cols}
       FROM tww_od.re_maintenance_event_wastewater_structure mw
-         INNER JOIN tww_app.vw_tww_channel ch ON ch.obj_id = mw.fk_wastewater_structure
+         INNER JOIN tww_app.vw_tww_channel ch ON ch.ch_obj_id = mw.fk_wastewater_structure
```
ponceta, line 180: @sjib I did not look too deep but if a view is called channel, the channel attributes will not have a `ch_` prefix so in your cas it will be `obj_id` instead of `ch_obj_id`. Prefix are used to distinguish attributes coming from other elements.

datamodel/app/view/maintenance_views.py
```diff
@@ -131,7 +177,7 @@ def vw_tww_channel_maintenance(connection: psycopg.Connection, extra_definition:
...
```
ponceta, line 180:
```suggestion
         INNER JOIN tww_app.vw_tww_channel ch ON ch.obj_id = mw.fk_wastewater_structure
```

sjib:
@cymed thanks for all work on this!

Issue #653, opened 2025-07-03
Link to *_roles.sql not working anymore


sjib:
**Describe the bug**
_A clear and concise description of what the bug is._

Je souhaite tester une installation de TWW, mais les liens indiqués dans la documentation semblent erronés. En particulier, le point "2.3.1. Procédure" fait mention du fichier tww_vx.y.z_structure_and_demo_data.backup qui n'est a priori plus dans le repo github.

**To Reproduce**
Exact steps to reproduce the behavior:
1. Go to '[https://teksi.github.io/wastewater/installation-guide/database-initialization.html#process](https://teksi.github.io/wastewater/installation-guide/database-initialization.html#process)'
2. Click on '12_0_roles.sql' or '12_1_roles.sql'
3. See error - File not found

**Expected behavior**
_A clear and concise description of what you expected to happen._

Documentation that is up to date with datamodel

**Screenshots / data**
If applicable, add screenshots or data to help explain your problem.

**Additional context**

sjib:
- Cluster roles (12_0_roles.sql) have been moved to https://github.com/teksi/wastewater/blob/main/datamodel/app/tww_app_roles.sql
- What happened to 12_1_roles.sql @3nids @ponceta @cymed

cymed:
they were deleted in #625, as the creation of roles was altered to fit pum

sjib:
> they were deleted in [#625](https://github.com/teksi/wastewater/pull/625), as the creation of roles was altered to fit pum
So no need to setup user roles anymore?

PR #655, opened 2025-07-07
Fix_old_sql_roles_links

ponceta:
Fix old sql roles links in the documentation before reworking the whole thing for the new way of installing TEKSI modules.

FIx https://github.com/teksi/wastewater/issues/653


sjib:
@ponceta Why do we have this message? Anything broken here?

```

Invalid workflow file: .github/workflows/docs-builder.yml#L27
The workflow is not valid. .github/workflows/docs-builder.yml (Line: 27, Col: 11): Input release_tag is required, but not provided while calling.
```

PR #661, opened 2025-07-08
[CI] Pass secrets to called workflow

domi4484:


3nids (approved):

ponceta:
GitHub is a bit picky with having a TOKEN not clearly listed as used in the referenced workflow :

`The workflow is not valid. .github/workflows/release.yml (Line: 26, Col: 17): Invalid secret, TX_TOKEN is not defined in the referenced workflow.`

But it seems that we can inherit the secrets of the referenced workflow.

But in our case it is declared as an env and not a secret so I'm not sure it solves our case :

https://docs.github.com/en/actions/how-tos/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow

PR #663, opened 2025-07-08
add tww_is_primary to big views

cymed:
For usage in visualisations, it is beneficial to pass the information whether a point/linestring is part of the primary network


sjib (approved):
Looks good to me - any other than vw_tww_reach and vw_tww_wastewater_structure?

sjib:
We can think of if this information also should be displayed in the forms or not.

cymed:
> We can think of if this information also should be displayed in the forms or not.

It is meant primarily to effectively filter on overview map. I would not show it in the forms

sjib:
@cymed thanks a lot for this enhancement!

PR #669, opened 2025-07-09
[WIP] database specific roles via PUM

cymed:
This PR aims at having database specific roles via PUM. This is necessary for hosted environments where the DBAdmin wants to maintain one DB Cluster only, but give specific users access to specific databases


cymed:
Must be tackled separately on Oqtopus

sjib:
@cymed can you open or link an issue on TMMT for this?

PR #674, opened 2025-07-16
2025 07 16 update documentation export extra organisations

sjib:
- [x] Documentation for https://github.com/teksi/wastewater/pull/650
- [x] New comment added for dxf


sjib:
@cymed 

- Did you think already about how to adapt the organisation layer in the QGIS project? I think tww_local_extension should be listed somehow.
- How about the other flag tww_active?

organisation_tww_local_extension

cymed:
tww_active is a filter which organisations are in active use

tww_custom_extension defimes which organisations to export with SIA405 Base Abwasser

cymed:
Sorry, it is local not custom

sjib:
@cymed Ok to add both flags to the project form like this?
And only disply fk_dataowner / fk_provider if local_extension

grafik

grafik

As with VSA Organisations we do not want an fk_ added

cymed (approved):

PR #677, opened 2025-07-16
Add view file and standard relations to support simple document management

ponceta:
A relation to vw_file for vw_tww_reach and one for vw_tww_wastewater_structure.

Not polymorphic, not using DMS, works fine.

Next steps :

- Bring polymorphic relation back
- Make it work with DMS plugin

Good to know : 
On windows, if you reference a datamedia, you have to use unix styled / instead of windows native \ otherwise you will have a silent error postgreSQL side.
Some enhancements could be brought here too.


ponceta:
It looks easier to add the DMS back :

image

But I'm still struggling on polymorphic side :

image

PR #679, opened 2025-07-17
Create datamodel dumps again

cymed:
Mistakenly deleted


ponceta:
@cymed do we require `if: ${{ matrix.extension != '' }}` to make it work for the agxx datamodel dump?

cymed:
> @cymed do we require `if: ${{ matrix.extension != '' }}` to make it work for the agxx datamodel dump?

I don't think we need a separate schemaspy for agxx

PR #684, opened 2025-07-24
Add option to filter out NULL values on import

cymed:
This PR adds a flag on import whether or not NULL entries should be passed from the intermediate schema to the od schema or not. The flag can be reached from the cmd.
When the filter_nulls flag is used, all attributes that are NULL are not passed to the od schema. This allows preventing an Import from overwriting existing data in cases where the import dataset is partial (i.e.due to the sender not being responsible for a certain field)


sjib (changes requested):
unexpected keyword argument 'xtf_file_output'

sjib:
Suggestion text adaption: Save logs next to import file

As we have no output on an import

sjib (approved):
Plugin functionality seems to be ok now.

PR #688, opened 2025-07-29
add log card view

cymed:
adds wastewater node identifier and situation geometry to log card table


sjib:
@cymed Good idea. Can you also add whether it is a PAA node?

cymed:
Not touching the plugin in this PR. Problem with qgis latest build @ponceta?

ponceta (approved):

PR #692, opened 2025-07-31
fix vw_tww_catchment_area_totals.sql

cymed:
The previous structure needed a self-reference on main log cards. This PR fixes that


urskaufmann (approved):
works

PR #698, opened 2025-08-05
Replace hardcoded 2056 SRID in catchment_area views with variable

w0pr:
contributes to https://github.com/teksi/wastewater/issues/586


ponceta:
Thank you for this nice contribution!

PR #699, opened 2025-08-05
fix datamodel-create-dumps.yml

3nids:


3nids:
@cymed I guess next one if for you ;) 

`TypeError: Hook.load_modification() got an unexpected keyword argument 'modification'`

3nids:
@cymed next one: FileNotFoundError: The file modification_repos\tww2ag6496\config.yaml does not exist.

PR #703, opened 2025-08-06
Divide actions

ponceta:
This is a draft and it can surely be better written but I think if we can run separate jobs for datamodel dumps creations and schemaspy, it could be for the best in terms of stability/readability.


3nids:
I would not recommend doing this as it increases maintenance effort, duplicates the code (and burns more feel).

If the concern if schemaspy, we could consider still continuing the workflow if the step fails.

PR #705, opened 2025-08-06
docker compose up

3nids:


ponceta (approved):

ponceta:
So `docker compose up` for persistent services and `docker compose exec` for executing scripts?

sjib:
May be this could help to understand?

https://stackoverflow.com/questions/65100572/docker-compose-difference-run-exec-and-what-happens-to-the-layers

and/or

https://stackoverflow.com/questions/62657762/docker-compose-up-versus-docker-compose-run-why-the-latter-seems-not-to-start-t

PR #713, opened 2025-08-12
rename measurement_result time and value fields

cymed:
`time` and `value` are not yet reserved in Postgres, but in SQL:2023


sjib:
`TypeError: 'time' is an invalid keyword argument for measurement_result`

sjib:
@cymed Why does the CI still complain - even with your rename?

plugin/teksi_wastewater/interlis/interlis_model_mapping/interlis_importer_to_intermediate_schema.py
```diff
@@ -1821,8 +1821,8 @@ def _import_messresultat(self):
                 ),
                 measuring_duration=row.messdauer,
                 remark=row.bemerkung,
-                time=row.zeit,
-                value=row.wert,
+                time_point=row.zeit,
```
sjib, line 1824: @cymed Can you add a comment on this - else we will not remember, why this was adapted: e.g. rename from time to time_point and from value to measurement_value as xxx implemented time and value as SQL key words - prevent future conflict on PostgresSQL

ponceta (approved):

PR #721, opened 2025-08-19
2025-08-19-harmonize-progress_done-on-interlis-export

sjib:
- adapt self._progress_done


sjib:
@urskaufmann Can you check if this solves your issue?
Is there a difference whether you export labels or not?

ponceta (approved):

Issue #724, opened 2025-08-20
DSS Import in tww2025.0.2rc5: AttributeError: 'int' object has no attribute 't_ili_tid' - rueckstausicherung


urskaufmann:
**Describe the bug**

Try to import the DSS.XTF, which I have exported with tww2025.0.2rc4 into new release rc5
and get this error:

An error has occurred while executing Python code: 

AttributeError: 'int' object has no attribute 't_ili_tid' 
Traceback (most recent call last):
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\teksi_wastewater_plugin.py", line 677, in actionImportClicked
    self.interlisImporterExporter.action_import()
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\interlis\gui\interlis_importer_exporter_gui.py", line 93, in action_import
    raise exception
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\interlis\gui\interlis_importer_exporter_gui.py", line 71, in action_import
    self.interlis_importer_exporter.interlis_import(
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\interlis\interlis_importer_exporter.py", line 180, in interlis_import
    raise exception
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\interlis\interlis_importer_exporter.py", line 134, in interlis_import
    tww_session = self._import_from_intermediate_schema(import_model)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users/KAU/AppData/Roaming/QGIS/QGIS3\profiles\tww_en_dev_2025-08-20/python/plugins\teksi_wastewater\interlis\interlis_importer_exporter.py", line 671, in _import_from_intermediate_schema
    interlisImporterToIntermediateSchema.tww_import(skip_closing_tww_session=True)
[...]

PR #725, opened 2025-08-20
20250820-fix-import-rueckstausicherung

sjib:
- correct relations in def _import_rueckstausicherung
- should solve https://github.com/teksi/wastewater/issues/724


sjib:
@ponceta Any hesitations to merge? Would be nice to have this in the next release also.

Issue #722, opened 2025-08-19
vw_tww_overflow.overflow_type is no more Enumeration


urskaufmann:
**Describe the bug**
overflow_type was in the old project Enumeration. Enumeration is know with 2025.0.2rc4 greyed out. It's no more possible to save a new record with this view.

**Desktop (please complete the following information):**
 - TWW version 2025.0.2rc4
 - QGIS Version 3.40.6
 - OS win10

urskaufmann:
In 2024.0.5 was the script

       CASE
            WHEN leapingweir.obj_id IS NOT NULL THEN 'leapingweir'::tww_app.overflow_type
            WHEN prank_weir.obj_id IS NOT NULL THEN 'prank_weir'::tww_app.overflow_type
            WHEN pump.obj_id IS NOT NULL THEN 'pump'::tww_app.overflow_type
            ELSE 'unknown'::tww_app.overflow_type
        END AS overflow_type,

in 2025.0.2 is the script

         CASE
            WHEN lw.obj_id IS NOT NULL THEN 'leapingweir'::text
            WHEN pw.obj_id IS NOT NULL THEN 'prank_weir'::text
            WHEN pu.obj_id IS NOT NULL THEN 'pump'::text
            ELSE 'unknown'::text
        END AS overflow_type,

Do I have to change the widget type of overflow_type from Enumeration to Value Map and add the values manually in the map?

Are there other enumerations, that have changed (or will they change?)?

PR #732, opened 2025-08-22
Update of QGIS project

ponceta:
Update of QGIS project by @urskaufmann


ponceta:
@urskaufmann here is the last project, once merged it will provide translations strings to transifex

sjib:
And the documentation of the changes:

 

loaded layers:

- vw_tww_channel

- vw_tww_channel_maintenance

- vw_tww_ws_maintenance

 

Rename layer-group Inspection to Examination-Maintenance

 

group  structure parts in group wastewater_structure doubled in group for "manholes" and group for channels.

all layers structure_parts 

- show fk_wastewater_structure

- fk_wastewater_structure as Relation Reference, not NULL enforced, on map identification (because a structure part can not exist without structure!)

load layers in group structure part manhole

- vw_electric_equipement and vl *_kind

- vw_electromechanical_equipement and vl *_kind

 

loaded layers (dssmini_structure_parts)

- vw_backflow_prevention   and vl *_kind

- vw_solids_retention and vl *_kind

- vw_tank_cleaning and vl *_kind

- vw_tank_emptying and vl *_kind

Layers are only visible in vw_tww_wastewater_structure depending on the special_structure.function

 

structure parts with nicer FormWindow, similar to all other Structure-Parts-FormWindows.

 

vw_tww_reach: all numeric values with constraint expression in range-definition

 

Special Relations:

- backflow-prevention. fk_pump and fk_trottle_shut_off_unit configured as Relation, only visible in AttributForm if kind is certain value

- tank-Emptying.fk_overflow and fk_trottle_shut_off_unit configured as Value_Relation. Could work like that, because there are not as many overflow / trottle_shut_off_units, so the performance should not be a problem. Only visible in AttributeForm if kind is certain value.

- trottle_shut_off_unit.fk_overflow configured as relation. only visible in AttributForm if kind is certain value

 

measures: order of attributes in AttributForm and attributetable as in Massnahmentabelle Kt Bern. But when exporting aus csv, this order has no meaning. Is it possible to do the csv-export in the order of the attribut-table? Or to create the view in this order, so this export has the structure we need?

 

vw_individual_surface: symbology without represent_value (I hoped, the symbology can also be translated with qm-file). Without automatic translation, this represent_value is not a good idea.

 

loaded layer measuring_point and 2 vl_measuring_point_*  (for dssmini logcard measuring). In Layergroup Wastewater_Structure. I think not to load the other measuring-tables because they are not necessary for log_cards.

loaded layer: vl_small_treatment_plant and 2 vl: is just a polygon-geometry, no point-geometry. Symbols are not possible. To be discussed

 

catchment_area: symbology without represent_value

 

vw_tww_wastewater_structure: Tab Maintenace_Event was changed. For me, it works not like this. Changed so its equal to vw_tww_reach.

 

vw_tww_overflow.overflow_type with ValueMap (multilingual), not Enumeration.

 

descriptions for fields with vl (hope I got all)

 

Styles *AG6496 not changed!

 

if Layer has more than one style, the configured style has now a prefix 0- in the name. This is the layer, where you can copy field. or attribut-form definitions to other layers. This layer must be choosen in the tww-project. This layer will then be translated.

 

To discuss:

we have now vw_tww_channel in the project (not editable). Can we therefor delete vw_channel (editable)?

sjib:
> descriptions for fields with vl (hope I got all)

@ponceta I think this needs a new datamodel release with https://github.com/teksi/wastewater/pull/734  to test.

sjib (changes requested):
see above

urskaufmann:
the configuration of vw_tww_channel_maintenance and vw_tww_wws_maintenance is not up to date because the fieldnames will change (the fieldnames in this two views for maintenance-fields should be similar). vw_tww_channel_maintenance is editable, vw_tww_wws_maintenance is no not editable.

sjib:
> the configuration of vw_tww_channel_maintenance and vw_tww_wws_maintenance is not up to date because the fieldnames will change (the fieldnames in this two views should be similar for maintenance-fields). vw_tww_channel_maintenance is editable, vw_tww_wws_maintenance is now not editable.

Are they correct here: https://github.com/teksi/wastewater/pull/733
Should this be merged first?

urskaufmann:
I need a database with the corrected fieldnames in vw_tww_wws_maintenance and vw_tww_channel_maintenance

sjib:
solves https://github.com/teksi/wastewater/issues/722

PR #734, opened 2025-08-25
reformat 2025.2 changelogs 61-64

cymed:
Due to the changelogs having a high number of lines, the generated log file in CI is truncated. This PR deletes unneccesary line breaks to prevent that


sjib:
I @sjib will work on my code to produce this

sjib:
You can merge.

PR #735, opened 2025-08-25
Fix create_app.py

ponceta:
Fix create_app.py


datamodel/app/create_app.py
```diff
@@ -431,15 +431,13 @@ def parse_variables(self, variables: dict) -> dict:
     parser.add_argument(
         "-a",
         "--modification_agxx",
-        type=bool,
```
ponceta, line 434: I don't understand why CI had not detected that it is not valid with python 12.

PR #754, opened 2025-09-10
Parse Obj ID into Foreign Key errors

cymed:
This PR passes the obj_id instead of the t_id on foreign key errors on export


sjib:
@cymed cooles Feature - das ist hilfreich!

domi4484 (approved):

PR #748, opened 2025-09-05
20250905 interlis export add 250 scale

sjib:
Solves https://github.com/teksi/wastewater/issues/747

Issue #747, opened 2025-09-05
Add 1:250 scale for label-export


urskaufmann:
Export of labels with scale_network_plan = 500(Werkplan) is not my preferred scale. I want a scale of 1:250 for network plan labels. I think an easy and working solution is to add both scales to the export-definition window, so the user can choose between 250 and 500. Of course it makes no seens to choose both options.

I think it's just 2 new lines in extractlabels_interlis.py

    tr = partial(QCoreApplication.translate, "ExtractlabelsInterlisAlgorithm")
    AVAILABLE_SCALE_PIPELINE_REGISTRY_1_1000 = "Leitungskataster"
    **AVAILABLE_SCALE_NETWORK_PLAN_1_250 = "Werkplan"**
    AVAILABLE_SCALE_NETWORK_PLAN_1_500 = "Werkplan"
    AVAILABLE_SCALE_OVERVIEWMAP_1_10000 = "Uebersichtsplan.UeP10"
    AVAILABLE_SCALE_OVERVIEWMAP_1_5000 = "Uebersichtsplan.UeP5"
    AVAILABLE_SCALE_OVERVIEWMAP_1_2000 = "Uebersichtsplan.UeP2"
    AVAILABLE_SCALES = [
        # ili key, display name, scale value
        (
            AVAILABLE_SCALE_PIPELINE_REGISTRY_1_1000,
            tr("Leitungskataster"),
            1000,
        ),  # TODO: check scale ?
        **(AVAILABLE_SCALE_NETWORK_PLAN_1_250, tr("Werkplan"), 250),  # TODO: check scale ?**
        (AVAILABLE_SCALE_NETWORK_PLAN_1_500, tr("Werkplan"), 500),  # TODO: check scale ?
        (AVAILABLE_SCALE_OVERVIEWMAP_1_10000, tr("Uebersichtsplan 1:10000"), 10000),
        (AVAILABLE_SCALE_OVERVIEWMAP_1_5000, tr("Uebersichtsplan 1:5000"), 5000),
        (AVAILABLE_SCALE_OVERVIEWMAP_1_2000, tr("Uebersichtsplan 1:2000"), 2000),

urskaufmann:
it works. Export with 1:250 and 2000 or 1:500 and 2000 is ok. Even 1:250, 500 and 2000 works, but makes no sense, because you can not distinguish which Werkplan is 250 and which is 500.

PR #756, opened 2025-09-16
20250909_fix_export_scale_1_250

sjib:
Trying to fix export scale featrure - as it seems not to work as expected (https://github.com/teksi/wastewater/issues/747#issuecomment-3272080820) with the adaptions of https://github.com/teksi/wastewater/pull/748
- Adapt scale


sjib:
Seems to work:

```
			<DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung_Text TID="ch000000RX005219">
				<TextPos>
					<COORD>
						<C1>2647320.257</C1>
						<C2>1141646.305</C2>
					</COORD>
				</TextPos>
				<TextOri>25.3</TextOri>
				<TextHAli>Left</TextHAli>
				<TextVAli>Bottom</TextVAli>
				<Plantyp>Werkplan.250</Plantyp>
				<Textinhalt>2000/1379 CS 1989 3870.15 73.6‰  </Textinhalt>
				<HaltungRef REF="ch080qwzHA000023"/>
			</DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung_Text>
```
and

```
			<DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung_Text TID="ch000000RX005372">
				<TextPos>
					<COORD>
						<C1>2647317.419</C1>
						<C2>1141640.294</C2>
					</COORD>
				</TextPos>
				<TextOri>25.3</TextOri>
				<TextHAli>Left</TextHAli>
				<TextVAli>Bottom</TextVAli>
				<Plantyp>Werkplan.500</Plantyp>
				<Textinhalt>2000/1379 CS 1989 3870.15 73.6‰  </Textinhalt>
				<HaltungRef REF="ch080qwzHA000023"/>
			</DSS_2020_1_LV95.Siedlungsentwaesserung.Haltung_Text>
```
if I select both export options. But we cannot work like this, as INTERLIS only knows `Werkplan`

- [ ] So we have to find out, how only Werkplan can be used - check how Leitungskataster was matched.

sjib:
grafik

sjib:
@cymed I now made the fix in the output of the xtf file by replacing Werkplan.xxx by Werkplan. All other approaches were too complicated. 

The only small thing is that you should not select both Werkplan scales at the same time - as you will get duplicated values and will not be able to sort out which scale is which.

Can you check if it also works with AG64/96 extension?

urskaufmann (approved):
Thanks Stefan

PR #801, opened 2025-10-27
Fix #797

### LINKED ISSUE #797: SWMM Export - Rain Gages IDs are identical for separative catchments
When exporting to SWMM, raingages are automatically created for catchments. For every catchment with two connections (separative system) in TEKSI, two catchments and two raingages are created. the catchments are named differently, but **the raingages have an identical name**, which gives an error in simulation

**To Reproduce**
Exact steps to reproduce the behavior:
1. Export to SWMM with separated system catchments.
2. Chose the default_qgep_swmm_parameters.
3. Start a simulation.
4. Error comes up.

**Expected behavior**
A different rain gage is created for every catchment.

**Desktop (please complete the following information):**
 - TWW version 
 - QGIS Version 3.4
 - OS Windows

cymed:
Fixes #797

> Think we should replace the name in the views
>
> add ca.state in the name.
>
>
> Image

 _Originally posted by @TriformMoritz in [#797](https://github.com/teksi/wastewater/issues/797#issuecomment-3446654455)_

### DIFF
--- a/datamodel/app/view/swmm_views/12_vw_swmm_raingages.sql
+++ b/datamodel/app/view/swmm_views/12_vw_swmm_raingages.sql
@@ -1,7 +1,7 @@
 -- Creates a default raingage for each subcatchment
 CREATE OR REPLACE VIEW tww_app.swmm_vw_raingages AS
 SELECT
-  ('raingage@' || replace(ca.obj_id, ' ', '_') '_',ca.state)::varchar as Name,
+  ('raingage@' , replace(ca.obj_id, ' ', '_'), '_', ca.state)::varchar as Name,
   'INTENSITY'::varchar as Format,
   '0:15'::varchar as Interval,
   '1.0'::varchar as SCF,


cymed:
Either copy-paste the change into your db or use the artifact from the git action

ponceta:
@TriformMoritz Here is how :

https://github.com/user-attachments/assets/b5c216ee-5012-4db0-ab0c-8d8c8d21c7e5

We have to put this somewhere on the doc, it will be usefull for everyone!

In the version tab, click on show developpement branches.

You should also be able to tick the demo data so you can test it directly.

This will still have the not created views issue btw. https://github.com/opengisch/oqtopus/issues/83

TriformMoritz:
Thank you, that is nice.
But for testing:
-If we have the same error with the demo data, why didn't anyone control it? **Testing should be part of any new fonction in TEKSI (or QGEP). For me, it is the most important part.** I have the feeling SWMM export was never tested.
-I want to test it with real data that I am using. If I use demo data, I am not sure it will work.

TriformMoritz:
Tried to execute the sql directly, but I got an error:

ERROR:  erreur de syntaxe sur ou près de « { »
LINE 10: ... st_centroid(perimeter_geometry)::geometry(Point, {SRID}) as...
                                                              ^ 

ERREUR:  erreur de syntaxe sur ou près de « { »
SQL state: 42601
Character: 478

And replacing {SRID} with 2056 gives:

ERROR:  erreur de syntaxe sur ou près de « { »
LINE 10: ... st_centroid(perimeter_geometry)::geometry(Point, {SRID}) as...
                                                              ^ 
ERROR:  le type « replace » n'existe pas
LINE 4:   ('raingage@' || replace(ca.obj_id, ' ', '_') '_',ca.state)...
                          ^ 

ERREUR:  le type « replace » n'existe pas
SQL state: 42704
Character: 130

So not possible to execute it.

TriformMoritz:
To do that, would take me about 2-3 hours, so I made a script on the text which corrected the error (not a good way, but much easier). I will wait for the upgrade tool to work in TMMT to test.

Issue #697, opened 2025-08-04
Not all Views are created when deploying via oqtopus (COMMIT missing?)


w0pr:
**Describe the bug**
Deploying prerelease 2025.0.2rc3 via oqtopus 1.1.19, the views created at this point in create_app.py https://github.com/teksi/wastewater/blob/fac93a8c9921769364d18698861b746db142e732/datamodel/app/create_app.py#L247 never show up in the database.

Adding "self.execute('COMMIT;')" after creation of the views (or in run_sql() even) make them show up after oqtopus is finished.

I have no clue about PUM, there a mentions of never calling COMMIT in a hook and so on in the PUM docs, but there must be a way to fix this.

sjib:
I can confirm - also with TMMT 0.1.20

This causes then also problems with the project - as layers cannot be created

![Image](https://github.com/user-attachments/assets/c5c6a8f7-a6ee-417f-9067-7e62af1fe6ea)
![Image](https://github.com/user-attachments/assets/f584fa00-9f48-4bb3-9a4b-83ecefea3698)
![Image](https://github.com/user-attachments/assets/4f91e8dc-9925-4b72-984a-ba3e0ebd2752)

@ponceta Can you feed this in to TMMT?

w0pr:
Somehow also the permissions are messed up, I had to run the "old" 12_1_roles.sql script to make the project work after deploying 2025.0.2rc3 via oqtopus 1.20.

ponceta:
Will be fixed in next rc. I merge a few things and do the rc today

ponceta:
New RC is available and should fix this :

https://github.com/teksi/wastewater/releases/tag/2025.0.2rc4

w0pr:
> New RC is available and should fix this :
> 
> https://github.com/teksi/wastewater/releases/tag/2025.0.2rc4
> 
> I let you test it with latest octopus 0.1.20 and close if fixed for you.

testing with oqtopus 0.1.20:

<img width="1086" height="572" alt="Image" src="https://github.com/user-attachments/assets/95a54925-e3ff-46e4-a5d2-e61744677aa6" />

<img width="675" height="179" alt="Image" src="https://github.com/user-attachments/assets/726b8ee2-5fe4-4273-ba84-dc1f1b6ce38a" />


Upgraded PUM from QGIS python console:
`import pip; pip.main(["install", "--upgrade", "pum"])`
but does not seem to help...still get the same error after "Load package task finished".

```
[...]

PR #808, opened 2025-10-31
Fix app creation statements order for oQtopus

ponceta:
Investigating to fix https://github.com/teksi/wastewater/issues/697

Apparently, statement order has a result on what is pushed in the database from QGIS and oqtopus and it may be related to the fact that some statements are not committed to the database.

See https://github.com/teksi/wastewater/pull/811 for investigating invalid commit mechanics.

### DIFF
--- a/datamodel/app/create_app.py
+++ b/datamodel/app/create_app.py
@@ -306,7 +306,6 @@ Running modification {modification.get('id')}
         # run post_all
         self.run_sql_files_in_folder(self.cwd / "post_all")
 
-
     @staticmethod
     def load_yaml(file: Path) -> dict[str]:
         """Safely loads a YAML file and ensures it returns a dictionary."""


ponceta:
More on that in the upcoming week. It solves the issue but now we have to understand why and fix it for every future module.

datamodel/app/create_app.py
```diff
@@ -95,6 +95,22 @@ def run_hook(
...
+            "view/swmm_views",
+            "view/network",
+        ]
+
+        for directory in sql_directories:
+            abs_dir = self.cwd / directory
+            self.run_sql_files_in_folder(abs_dir)
+
+        # run post_all
+        self.run_sql_files_in_folder(self.cwd / "post_all")
```
3nids: this should stay at the end I believe?

cymed: Yes, to re-set the triggers etc.

PR #823, opened 2025-11-12
Fallback value for damage_channel geometry

cymed:
This PR adds a fallback value for damage_channel geometry


ponceta:
AS ST_Length(base.ch_progression2d_geometry) can be 0 if ch_progression2d_geometry is NULL, we should prevent somehow the  `division by 0`.

ponceta (approved):
Great! Thank you!

PR #835, opened 2025-11-18
activate extension testing

3nids:
This reverts commit 839b4bd82f4a6b12ed48ff7826f56370bef830a0.

@cymed as discussed in #832, this activate extension testing (which were not before). They are failing so it needs a bit of attention:

image


ponceta:
Thank you @3nids and @cymed for this which will ease the monitoring of compatibility between TWW and its AG relatives.

PR #852, opened 2025-11-28
2025-11-27-new-project-tww

sjib:
Changes by @urskaufmann :
-	Some symbols of vw_tww_maintenance_channel are corrected
grafik

- The symbols of vw_tww_maintenance_channel have now Symbol Levels 

grafik

   Symbol Levels are important for this view, because there can be many symbols at the same place. -> remark in documentation.
-	vw_tww_maintenance_event. Field maintenance_event_type has enforced constraint "maintenance_event_type" is not 'unknown' -
-	Reuse last entered value is disabled for all fk_dataowner and fk_provider – so it's clear they are defined with the default_values.


sjib:
- [ ] Not yet configured: vw_tww_catchment_area_totals in Layergroup Log Card!

PR #858, opened 2025-11-28
2025 11 28 hoehenbreitenverhaeltnis model patch available

sjib:
Preparation to change import code and tests to Patch Change of VSA end of November


sjib:
- [ ] Re-run checks as soon as patch Change is online in model repo www.vsa.ch/models

sjib:
@w0pr Any idea what to change?

sjib:
@w0pr Thanks for the hint. Pre-commit has taken this out.

Can you test if it now works as expected? 
The latest VSA Models you can find here: https://vsa.ch/models/2020_1

PR #859, opened 2025-11-28
improve performance of mvw_tww_channel

cymed:
By using a Common Table Expressions (CTE) to aggregate the reaches, the performance of the materialized view improves drastically

https://www.postgresql.org/docs/current/queries-with.html

### DIFF
--- a/datamodel/app/view/maintenance_views.py
+++ b/datamodel/app/view/maintenance_views.py
@@ -137,35 +137,6 @@ def mvw_tww_channel(
                 "fk_main_cover",
             ],
         ),
-        ch_cols_grp=select_columns(
-            connection=connection,
-            table_schema="tww_od",
-            table_name="channel",
-            table_alias="ch",
-            remove_pkey=False,
-            indent=4,
-            skip_columns=[],
-        ),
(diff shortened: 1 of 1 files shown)


datamodel/app/view/maintenance_views.py
```diff
@@ -51,30 +59,39 @@ def mvw_tww_channel(
...
+            SELECT
+                cg.obj_id,
+                rs.fk_reach_point_from,
+                re.fk_reach_point_to
+            FROM channel_geometries cg
+            JOIN reach_start_end_points rs ON rs.fk_wastewater_structure = cg.obj_id AND st_equals(rs.reach_start_point, cg.channel_start_point)
+            JOIN reach_start_end_points re ON re.fk_wastewater_structure = cg.obj_id AND st_equals(re.reach_end_point, cg.channel_end_point)
+        ), re_agg as(
+        select ne.fk_wastewater_structure,
+            st_multi(st_forcecurve(st_linemerge(st_collect(re.progression2d_geometry))))::geometry(MultiCurve,2056) AS progression2d_geometry,
```
w0pr: Should use {{srid}} param instead of 2056.

PR #863, opened 2025-12-01
Add option to disable reach level sync

cymed:
This PR adds a bool to define whether or not we overwrite the reach level when altering the 3d geometry

### DIFF
--- a/docs/en/user-guide/digitizing/digitizingchannels.rst
+++ b/docs/en/user-guide/digitizing/digitizingchannels.rst
@@ -86,7 +86,7 @@ Geometry synchronization
 
 When inserting a new reach using the wizard, the start and end points of the line geometry determine the level of the corresponding reach points (as shown in the pictures above).
 These values can be overwritten before saving.
-Upon saving, the level values `rp_from_level` and `rp_to_level` are used to define The Z value of the first and last vertex of a reach.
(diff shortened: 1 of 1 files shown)


cymed:
The field `tww_update_lvl_by_geom` is a boolean that is false per default. It is populated on update by the settings variable `@tww_update_lvl_by geom`. If the value is `true`, the Z coordinate of the `progession3d_geometry`'s first/last vertex is used to overwrite the `reach_point.level`. If false, it does not.

> If I now pull the middle-point on 3d-point2, the existing level (p1-level) is overwriten with p2-level. I get now z1,p2-level,z2. In this case, I would be happy about a warning/level-reset-option. 

Intermediate vertices are not accessed by the code. The code does the following:

1. Extract the Z value of the start node
2. Compare with the corresponding `NEW.reach_point.level`. If they are the same, skip and don't alter anything
3. If the `OLD.reach_point.level` and the `NEW.reach_point.level` are distinct, then overwrite the geometry's Z
4. If not, overwrite the `NEW.reach_point.level` if `tww_update_lvl_by_geom` is true, the geometry's Z is not NULL or 0 and the geometry's Z has changed

The only change introduced was the check of `tww_update_lvl_by_geom`

PR #877, opened 2025-12-12
[WIP] enable unittest that require demodata

cymed:
as a check for #612 if the base was broken already


cymed:
Status: apart from the missing demo data, the relations are now correct in this PR. I guess the tests were built based on Arbon @ponceta ?

PR #882, opened 2025-12-16
refresh materialized views concurrently

### LINKED ISSUE #868: SQL refresh network topology makes QGIS crashe if there are uncommited transactions
This implies that you have to commit your changes before refreshing the topology which is another issue that we'll have to fix.

Otherwise the uncommited transaction will make QGIS crash :

image

I open another issue for that.

cymed:
Fixes #868 by adding a CONCURRENTLY to the refreshing of the materialized views. 

From the Postgres Docs:

CONCURRENTLY

 Refresh the materialized view without locking out concurrent selects on the materialized view. Without this option a refresh which affects a lot of rows will tend to use fewer resources and complete more quickly, but could block other connections which are trying to read from the materialized view. This option may be faster in cases where a small number of rows are affected.

This option is only allowed if there is at least one UNIQUE index on the materialized view which uses only column names and includes all rows; that is, it must not be an expression index or include a WHERE clause.

 This option can only be used when the materialized view is already populated.

Even with this option only one REFRESH at a time may run against any one materialized view.


ponceta (changes requested):
As described

ponceta:
Some additional to that : 

When loading the tww project I get some postgis warnings : (which are not very explicit)

image

datamodel/app/view/maintenance_views.py
```diff
@@ -141,6 +140,13 @@ def mvw_tww_channel(
 
     matview_sql = psycopg.sql.SQL(matview_sql).format(srid=psycopg.sql.Literal(srid))
```
ponceta: There are several differences between the `vw_tww_wastewater_structure.py` and the `maintenance_views.py`

Are these differences required for any reasons?

cymed: What exactly are you referring to?

ponceta: the above statement for example.
`    matview_sql = psycopg.sql.SQL(matview_sql).format(srid=psycopg.sql.Literal(srid))`

cymed: Is an equivalent to https://github.com/teksi/wastewater/blob/fe3791e06e04c3429ff4220a36b2ae3115803f5c/datamodel/app/view/vw_tww_wastewater_structure.py#L218

ponceta:
pum check would be handy for sure ;)

cymed:
> pum check would be handy for sure ;)

we already use PUM in the tests. The problem is that we can run TMMT tests only after the artifacts were created.

sjib:
@cymed What's the status of this?

cymed:
CI passes, but there were issues with creating the materialized views in TMMT. I have not tested with the current TMMT release

cymed:
Just tested with TMMT version 1.4.2, pum version 1.7.1
The indexes were created

sjib:
@cymed Is this ready to be integrated in a pre-release?

sjib:
@ponceta Can you re-check?

sjib:
Please also check https://github.com/teksi/wastewater/pull/994

ponceta (approved):

PR #886, opened 2025-12-17
fix secret on release

3nids:


ponceta (approved):
Secrets handling is so much fun! 🥳

PR #909, opened 2026-01-20
handle pg2ili as an app schema and fix roles granting

3nids:
This PR modifies the handling of the pg2ili schema to have it as an application schema.
Permissions are now correctly handled by the pum config.

I also removed the post_all role handling which would cause an error if roles were not granted.
An option will be available in pum after doing an upgrade or recreating the app.


cymed (changes requested):
We must not change changelogs in 2025.0.2. 2025.0.2 was already published.  I'd prefer a `DROP SCHEMA IF EXISTS` in a separate changelog (2026.0.1)

PR #924, opened 2026-02-05
Drop schema if exists

ponceta:
DROP SCHEMA IF EXISTS pg2ili_abwasser  to be coherent with the changelog alteration between 2025.0.2 and 2026.0.1.


ponceta:
followup to #909

PR #932, opened 2026-02-13
Add info whether ws has log card to main view

cymed:
Add info whether ws has log card to main view


sjib (approved):
Locks good to me. Great addition.
@urskaufmann Does this fit you also? Any other ideas on that?

PR #943, opened 2026-03-05
Drop CI modification

cymed:
As we have other modifications, we do not need a designated CI modification anymore


cymed:
@sjib move to 2026.0?

ponceta (approved):

PR #944, opened 2026-03-09
Drop uuid-ossp

cymed:
uuid generation is native since PG 14

Edit : PG 14 by @ponceta 
ref https://www.postgresql.org/docs/current/functions-uuid.html


cymed:
> We have two ways, but in any case we do need changelogs.
> 
> The "clean way":
> * keep existing changelogs
> * new changelog with drop extension + alter fields
> 
> The "dirty or pragmatic" way (let's stick to pragmatic):
> * alter existing changelogs (as in the current state of this PR)
> * add new changelog to alter fields
> 
> I'd vote for the pragmatic approach. This removes the addition of an extension and this makes it more future proof.

Or we do both, so new dbs are created without the extension and existing ones can get rid of it

ponceta:
With the proposed alterations we will not create the extension anymore and we will drop it for all users if it exists when this delta is applied.

ponceta (approved):

Issue #225, opened 2024-04-28
dataowner, provider in vw_tww_reach


urskaufmann:
there is no attribute ws_fk_dataowner and ws_fk_provider.
there are the attributes for the reachpoints and for the reach (means vor the networkelement). But the more important (in my opinion) is the one of the wastewater_structure.  like that I can not manually set dataowner or provider for the wastewater_structure.
I hope I have got the actual datamodel (tww_dev_structure_with_value_lists.sql 09.04.2024 16:19)

ponceta:
- [ ] Unskip columns in : https://github.com/teksi/wastewater/blob/ef665f0d91483db83a38f86f10c467601efd484a/datamodel/app/view/vw_tww_reach.py#L124

sjib:
The question is, if we still have cases where the cadastre is done by one engineering company and the hydraulics by an other? Or different experts using TEKSI. And that we want to distinguish this between wastewater_structure (meanging the constructive part) and wastewater_network_elements (wastewater_node / reach as hydraulic parts)?

urskaufmann:
Im do not think it's good do set these attributs dataowner and provider through the nodes and reaches, just because we use there geometry in QGIS.
If I add a second wastewater-node and forget to set the dataowner -> will it then change the dataowner of the wastewater_structure or ask me, which one is correct?
We can do it like this, if the dataowner is set automatic in the connected classes as @ponceta describes. For me it's more logical, that normally the wastewater_structure is the class to define the dataowner and the other classes will inherit (if there is no entry).

I can not underwrite, that there will never be a case, where a node must have another dataowner than the wastewater_structure. If this case occures and there is no dataowner-field for this case in the view, you will have to go to postgres or load the table itself or something else not userfriendly to be able to change the value. If the fields are in the view, they must not be in the attribut-form, but I can easly set a new value with the field-calculator.

In vw_tww_wastewater_structure we have both fields (fk_dataowner = wastewater_structure.fk_dataowner, wn_fk_dataowner = wastewater_networkelement.fk_dataowner).

In vw_tww_reach we have fk_dataowner = wastewater_networkelement.fk_dataowner and should have also ws_fk_dataowner = wastewater_structure.fk_dataowner.
The other two reachpoint-fields (rp_from_/rp_to_fk_dataowner should be skiped, because I'm shure, there will never be a reachpoint with a dataowner that is not the same as the reach-dataowner...
All I wrote above for dataowner is similar to the provider-attribute.

[...]

PR #961, opened 2026-03-25
Urskaufmann doc changes in dataowner fields

urskaufmann:
I was struggling again over the missing "channel" ws_fk_dataowner-field in vw_tww_reach as mentoned in #225 .
The documentation how TWW works  in case of changing dataowner or provider was missing.
It shows, that TWW is not always very logical...


sjib (approved):
Description of what is happening is good. 
Further developments need a new issue.

ponceta (approved):

PR #970, opened 2026-04-08
Use QGIS stable

ponceta:
https://github.com/qgis/QGIS/issues/65732#issuecomment-4204013119


ponceta:
This fails because QT6 is not fully supported on the TWW plugin yet.

PR #976, opened 2026-04-17
Add standardoid as domain

cymed:
This PR enforces obj_id to be standardoid.

If there are non-conform standardoid in the existing data, the upgrade is blocked.


sjib (commented):
Nice enhancement that will help us that no non- conform obj_ids can be generated anymore.

What is the message if I add a dataset with a non-conform obj_id?
Do I get a message from postgres?
Or QGIS?
Also in pgadmin?

cymed:
It throws:

`ERROR:  value for domain tww_od.interlis_standardoid violates check constraint "interlis_standardoid_check"`

If we want more verbosity, we'd need a BEFORE TRIGGER like
```
CREATE OR REPLACE FUNCTION tww_app.check_interlis_oid()
RETURNS trigger AS $$
BEGIN
    IF length(NEW.obj_id) <> 16 THEN
        RAISE EXCEPTION
            'INTERLIS Standardoid must be exactly 16 characters (got %)',
            length(NEW.obj_id);
    END IF;

    IF NEW.obj_id !~ '^[A-Za-z0-9]+$' THEN
        RAISE EXCEPTION
            'INTERLIS Standardoid must be alphanumeric (got "%")',
            NEW.obj_id;
    END IF;

    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```
on all tables

sjib:
The only disadvantage of this is that we cannot translate it - I guess.
@ponceta would it need more information on how a Standardoid has to look like?

sjib:
@cymed What happens if someone wants to import non correct data in the database - will this be blocked?

cymed:
> @cymed What happens if someone wants to import non correct data in the database - will this be blocked?

yes

cymed:
Merge second of my open pr

ponceta (approved):

PR #983, opened 2026-04-28
Fix SWMM input creation

cymed:
- using TwwLayerManager to avoid translation issues
- using tww_is_primary on the db side

### DIFF
--- a/datamodel/app/view/swmm_views/02_vw_swmm_junctions.sql
+++ b/datamodel/app/view/swmm_views/02_vw_swmm_junctions.sql
@@ -19,7 +19,7 @@ SELECT
 		ELSE 'current'
 	END as state,
 	CASE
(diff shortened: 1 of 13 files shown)


tproduit:
Did you test all the functionnalities after your changes? The obj_id might be used to export selected object only.

sjib:
@cymed And don't we need the obj_id to re-connect results back to tww. What is the reason to eliminate?

cymed:
> @cymed And don't we need the obj_id to re-connect results back to tww. What is the reason to eliminate?

it caused problems on the creation of junctions

cymed:
Found problems elsewhere, reverting obj_id part.

ponceta:
@cymed @tproduit any additional commits required here?

cymed:
Ready for review from my side

sjib:
@tproduit Can you review again?

sjib:
@tproduit Can you review - if ok before 26.6 it can be moved to Milestone 2026.0

ponceta (approved):

PR #991, opened 2026-05-08
Fix AGXX import reach point coordinates

cymed:
There was an error with the Y coordinate of reach points


sjib (approved):
Looks logic to me to be corrected

PR #992, opened 2026-05-08
move refresh topology button to admin mode

cymed:
As long as the query runs so long, it is safer not to expose it for all users


ponceta:
The query is long? It takes usually 2-3 seconds by me?

cymed:
I have a db where it runs 15s

ponceta:
To check if it is still the case with your last addition.

cymed:
Still the case

cymed:
@sjib 2026.0?

sjib:
@ponceta @urskaufmann ok like this - please add your reviews

ponceta (approved):

ponceta:
Has to be documented somewhere :)

PR #994, opened 2026-05-13
Improvements on network refresh

cymed:
This PR adds the following improvements to the network refresh

- Raise timestamps usibg pg_notify (can be included into ui later)
- order by geometry distance by gist instead of postgis functions (uses indexes)
- refreshing the network materialized views does no longer update all materialized views, but only the network ones. The others are refreshed separately


sjib (commented):
Conception looks good to me. I cannot make a statement on the change in coding.

ponceta (approved):

PR #999, opened 2026-05-18
Import vsa organisation from internet if not present

### LINKED ISSUE #205: Organisations
We should ease the initialisation of organisations :

- [ ] Load latest organisations from VSA.
- [ ] Update organisations with latest organisations from VSA
- [ ] Ease the activation of organisations.
- [ ] Define default dataowner
- [ ] Define default provider

cymed:
This PR tries downloading the vsa organisations from the internet if there are no entries in `tww_od.organisation` and imports those before anything else


sjib:
@cymed 
Das kann man jetzt auch sehr einfach mit QGIS Bordmitteln machen:

grafik

cymed:
But it makes implementation in environments without QGIS more difficult

sjib:
I like the idea - only point to discuss if the VSA Organisations will get imported automatically when there are no organisations
@urskaufmann @TriformMoritz @ponceta @tproduit ... Any objections to this?

sjib:
@cymed Can you add also the documentation if ok?

TriformMoritz:
I think it would be very useful to import the VSA organisation with the database creation from the TMMT. Last time I tried it with the model baker. It did not work (some incomprehensible python error). 
So I just created two organisations. Which was also very complicated, because the "non active" organisations in the table are disabled by a query, so when you create a new line in the table, it simply disappears.

sjib:
> I think it would be very useful to import the VSA organisation with the database creation from the TMMT. Last time I tried it with the model baker. It did not work (some incomprehensible python error). So I just created two organisations. Which was also very complicated, because the "non active" organisations in the table are disabled by a query, so when you create a new line in the table, it simply disappears.

@TriformMoritz Why would you do that with ModelBaker - VSA Organisation is a regular VSA-DSS / SIA405 Abwasser 2020.1 INTERLIS Import!

You can turn off the filter on the table od_organisation

TriformMoritz:
I think I tried that, but will try it again (maybe it was my error!). Anyway, would be nice to have it in TMMT, same for the organisation prefix.

urskaufmann:
Good thing for me. I dont see the project, that will not need the organisations. And if there will be once the case, I will delete these organisations.
If possible: there should not be necessary to confirm the organisations at 90% of the import...

ponceta:
I like the idea but I would add a button for that instead of doing it on every import.

ponceta:
For now it could be an optional checkbox on the import window

cymed:
>  same for the organisation prefix.

See #1003

ponceta (approved):

PR #1002, opened 2026-05-22
use wastewater node identifiers in reach wizard identifier creation

cymed:
This pull request improves the reach digitizing workflow by adding client-side identifier generation in QGIS. The identifier is derived during digitizing from the snapping results of the start and end points. 

If snapping does not resolve to nodes, the logic falls back to the prefilled reach point object IDs allowing the workflow to remain functional for incomplete topology or reach-to-reach connections. If no valid values are available, the identifier defaults to reach.obj_id.

Uniqueness must be enforced via QGIS constraints, i.e. for parallel reaches.


cymed:
Self-tests worked

cymed:
@sjib This PR has been set as "ready for review" three weeks ago. Can we include it in 2026.0? Big Quality of Life upgrade

sjib:
Von mir aus ok

cymed:
Documentation pending

ponceta (approved):

PR #1003, opened 2026-05-26
add oid prefix to TMMT

cymed:
By adding a pum variable and a new app func, we can now automate oid prefix setting to TMMT


ponceta (approved):
Great addition, would require a small chapter in the upcoming how to install documentation.

ponceta (changes requested):
```
ERROR:  valeur trop longue pour le type character(8)
CONTEXT:  instruction SQL « INSERT INTO tww_sys.oid_prefixes (prefix, active)
        VALUES ('{oid_prefix}', TRUE) »
fonction PL/pgSQL inline_code_block, ligne 12 à instruction SQL 

ERREUR:  valeur trop longue pour le type character(8)
SQL state: 22001
```

ponceta (approved):

cymed:
pum does not yet support limitations via regex, see https://github.com/opengisch/pum/pull/194

cymed:
just noted the injection risk as long as pum cannot enforce regex limitations

ponceta:
Is there a way to enforce visually that the oid has to be 8 char long on the UI side? Does pum.yaml already allows field constraints?

ponceta:
```
Module operation error in 'install': SQL execution failed for the following code: SELECT tww_app.set_oid_prefix('ch123456');
 il n'existe aucune contrainte unique ou contrainte d'exclusion correspondant à la spécification ON CONFLICT
CONTEXT:  instruction SQL « INSERT INTO tww_sys.oid_prefixes (prefix, active)
    VALUES (v_oid_prefix, TRUE)
    ON CONFLICT (prefix)
    DO UPDATE SET active = TRUE »
fonction PL/pgSQL tww_app.set_oid_prefix(text,boolean), ligne 10 à instruction SQL
```

ponceta (changes requested):
Looks like it doesn't work on my pg instance

cymed:
works after adding unique constraint

ponceta (approved):

sjib:
@cymed Is there a check, that you can only add valid OID prefixes? And valid for Switzerland?

sjib:
@cymed Can you also adapt the documentation ?

ponceta:
> @cymed Is there a check, that you can only add valid OID prefixes? And valid for Switzerland?

I think that's what https://github.com/teksi/wastewater/pull/976 is about, but not quite exactly

sjib:
> > @cymed Is there a check, that you can only add valid OID prefixes? And valid for Switzerland?
> 
> I think that's what #976 is about, but not quite exactly

Yes, this is related - but I think this is too late, if I already have set a wrong prefix.

cymed:
> @cymed Is there a check, that you can only add valid OID prefixes? And valid for Switzerland?

See https://github.com/opengisch/pum/pull/194

sjib:
Mit regex: ^[\w\d]{8}$ sind beide prefixes gültig

 

de123456

ch123456

 

Also bisher keine Prüfung ob CH prefix

cymed:
If a german entity wants to, they can use TWW with a german standardoid

PR #1010, opened 2026-06-02
Versionized docs

cymed:
This PR introduces multi-version documentation support via CI. Previously, only a single “latest” version of the documentation was built and deployed.

Documentation is built per version:
- main → latest
- tags (YYYY.X.Y) → versioned directories

No docs for 2025.0.2 and previous are created, as the new rigidness of sphinx 9 would fail on their creation. It is to be decided if they are refactored or left aside.

Other Changes:
- Builds restore and update the existing gh-pages content instead of overwriting it
- A versions.json file is generated to drive a custom version selector

PR #1013, opened 2026-06-10
WIP : Fix installation documentation

ponceta:
The fix is currently in the wrong place.

We have currently an installation guide and an admin guide providing both hints for the database initialization.

It would be wise to simplify it.


cymed (approved):

cymed:
See also #1010 that fixes some things due to sphinx 9

PR #1050, opened 2026-06-29
Add chapter Obj_ID in TWW

sjib:
@urskaufmann Thanks for the text


urskaufmann (approved):
Thanks stefan

docs/en/admin-guide/db-production-readiness/index.rst
```diff
@@ -6,11 +6,39 @@ Prepare database for production
 OID Prefix Settings
 -------------------
 
-To prepare the database to be production ready change the active prefix in the tww_sys.oid_prefixes table to your project. If needed add your project prefix to that table first. Make sure only one prefix is set to active.
+To prepare the database to be production ready change the active demo prefix (ch000000) in the tww_sys.oid_prefixes table to your project prefix. If needed add your project prefix to that table first. Make sure only one prefix is set to active.
```
cymed, line 10: The OID Prefix can be set in TMMT from 2026.0.1 on

docs/en/admin-guide/db-production-readiness/index.rst
```diff
@@ -6,11 +6,39 @@ Prepare database for production
...
+
+These sequences store the next available sequential number for each class.
+
+What happens when new data is imported?
+* If the new data uses the same OID prefix as the OID prefix in the database, then the sequences must be updated. TWW does this automatically after the data import. The user does not need to do anything.
+* If the new data uses a different OID prefix than the selected OID prefix in the database, then the sequences do not need to be updated.
+
+What happens if the OID prefix in `tww_sys.oid_prefixes` is changed and data already exists in the database?
+* If the existing data already uses the new OID prefix, the sequences must be updated. TWW does this automatically after saving the new OID prefix.
+* If the existing data does not use the new OID prefix, the sequences must be reset so that the counter starts at 0 again with the new prefix. TWW does not perform this reset automatically.
```
cymed: There is a trigger that runs on UPDATE, so either when you alter the existing `active` or change the prefix of the active row.

cymed (changes requested):

docs/en/admin-guide/db-production-readiness/index.rst
sjib: @cymed So if I understand you right, this statement is not correct and these two cases do not have to be distinguished? 
_If the existing data does not use the new OID prefix, the sequences must be reset so that the counter starts at 0 again with the new prefix. TWW does not perform this reset automatically._

Since which release is the on UPDATE active?

cymed: Correct. I'd have to search for the PR

ponceta: https://github.com/teksi/wastewater/blame/a9584d58ac06f09c31b121c875714c2d6c829a55/datamodel/app/sql_functions/oid_functions.sql#L153

cymed: @ponceta not the part we we were discussing. Feature was added in #223, so already in 2024.0.3

sjib: @cymed @urskaufmann Ok, I have adapted the text

sjib:
@ponceta Should be ready to merge now

ponceta (approved):

PR #1063, opened 2026-07-02
Update wastewater node cache after updating wastewater_structure

### LINKED ISSUE #1062: snapping to recently added wastewater nodes does not work
**Describe the bug**
In the wizard, I cannot snap to wastewater nodes after creating the wastewater structure due to non-updated caches. Fix incoming.

cymed:
Fixes #1062 
Added the dependency from wastewater structure to wastewater node in qgis project


cymed:
@ponceta 2026.0 or 2026.0.2 ?

ponceta:
The earlier the better if we can fix this.

Can you link the issue?

cymed:
#1062 is linked

ponceta:
Shouldn't we have `vw_tww_wastewater_node` instead of `vw_wastewater_node` in 2026.X?

image


Or has a rollback occured on the TWW project?

cymed:
true - I do not know where that snuck in - will change

cymed:
touching the plugin too, as there were links to vw_wastewater_node

ponceta (approved):

PR #1064, opened 2026-07-02
List common editing mistakes in docs

cymed:
Necessary for formations

### DIFF
--- a/docs/en/user-guide/How-To/index.rst
+++ b/docs/en/user-guide/How-To/index.rst
@@ -315,7 +315,7 @@ Common mistakes
 Relying on the tww_app.vw_tww_wastewater_structure geometry
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 The layer tww_app.vw_tww_wastewater_structure create a point geometry from either the main wastewater node of a nodal wastewater structure or from its main cover.
-This is why the layer's geometry is 2D and we advise not to snap on the tww_app.vw_tww_wastewater_structure layer. 
+This is why the layer's geometry is 2D and we advise not to snap on the tww_app.vw_tww_wastewater_structure layer.
 Use tww_app.vw_cover or tww_app.vw_tww_wastewater_node instead, depending on the reason you snap.
 
 Ignoring reach points when splitting reaches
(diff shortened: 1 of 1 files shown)


docs/en/user-guide/How-To/index.rst
```diff
@@ -309,7 +312,43 @@ The Feature Attributes window for the overflow characteristic appears:
...
+The layer tww_app.vw_tww_wastewater_structure create a point geometry from either the main wastewater node of a nodal wastewater structure or from its main cover.
+This is why the layer's geometry is 2D and we advise not to snap on the tww_app.vw_tww_wastewater_structure layer.
+Use tww_app.vw_cover or tww_app.vw_tww_wastewater_node instead, depending on the reason you snap.
+
+Ignoring reach points when splitting reaches
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When a reach is split using the native tools, one of the resulting reaches keeps all the foreign keys of the old one while the other only keeps the channel obj_id.
+This means that the reach points of the old reach no longer correspond to the reach_point_from of the upper reach and the reach_point_to of the lower reach.
+This in turn means that the fk_reach_point of TV inspections might no longer reference the correct reach point.
+To keep consistency, one must load the the layers tww_od.reach and tww_od.reach_point, adapt the foreign keys and move the reach points to the correct location.
```
ponceta, line 329: Should we have something to do in the db to keep A and B the same and create new RP obj_id for the reach points created by the cutting?

Image

sjib: We have since long this idea of a Reach splitting Tool: https://github.com/QGEP/QGEP/issues/527

ponceta (approved):

docs/en/user-guide/How-To/index.rst
urskaufmann: the problem with the referencing of tv inspections can not be solved definitly, if there are changes in the network. TV inspection 1 is based on a network version 1, that has to be corrected (network version 2) because of the tv1-results.  tv inspection 2 is based on network version 2. After this you have to rebuild a part of the channel because of damages -> network version 3.  
It will probably no more be possible to reference tv1 with network3.
It will be always a good idea, to save the tv-inspections as independent geometry (the way the camera moved).
TWW can not solve the problem, its a dss-kek-problem.
But the hint in the docu about this problem is helpful

anyway, the splitting tool would be a helpful thing (choose if channel is also splitted, calculation of levels at the split-point, choose if there is a node at the split-point).

docs/en/user-guide/How-To/index.rst
```diff
@@ -309,7 +312,43 @@ The Feature Attributes window for the overflow characteristic appears:
...
+It is planned to check for database locks before firing the corresponding processes, but this is not implemented yet.
+
+Snapping fails on freshly created wastewater nodes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When creating a new wastewater structure, the database creates the corresponding wastewater node.
+The TWW wizard, however, does not refresh its memory of wastewater nodes to snap to, as it was created indirectly.
+Save, exit the editing mode, re-enter editing mode to snap on the entry.
+
+..versionadded: 2026.0.1
+ this issue was fixed in 2026.0.1
```
sjib: This issue was fixed in 2026.0.1

docs/en/user-guide/How-To/index.rst
cymed: I once started a reach splitting tool, but could not proceed on voluntary hours. It would be a good util to discuss in the user group

cymed: The problem cannot be solved on the db side in a stable way, as th db does not know if the newly generated reach is upstream or downstream

ponceta (approved):

PR #1065, opened 2026-07-03
Provide / fix translations for the plugin

ponceta:
Provide / fix translations for the plugin


sjib:
@ponceta Are those adaptions we have to integrate and document in the Developers Guide?

ponceta:
We hade to do more adaptions to fix https://github.com/teksi/wastewater/issues/1066 and these will have to be documented because the translation system set in https://github.com/teksi/wastewater/blob/main/plugin/teksi_wastewater/interlis/processing_algs/extractlabels_interlis.py does not work.

PR #1093, opened 2026-07-20
2026 07 20 add italien to actions

sjib:

PR #1111, opened 2026-08-04
Use makefile to build the docs

domi4484:


cymed:
See tests in #1093. #1093 should be superseded by this

ponceta:
Translations file are here and everything looks good but if you look at the artefact it is not translated :(

domi4484:
@ponceta @cymed The issue was an incorrect relative path in conf.py. If you download the artifact from here you will see the translations in place in the html: https://github.com/teksi/wastewater/actions/runs/30898123441?pr=1111

cymed (approved):

cymed:
Translations are there, but I just saw that translated paths are now in gh-pages/{lang}, while docs expect gh_pages/{version}/lang

ponceta:
Same feedback here.

PR #1118, opened 2026-08-07
Add italian for docs

ponceta:
Add italian for docs

image


ponceta:
@sjib as a follow up to this we will have to publish updated glossaries for Italian

ponceta:
Now we need to learn italian or find some italian translators!

sjib:
> Now we need to learn italian or find some italian translators!

I am working on it. We will have a new member hopefully soon from Italy

PR #1144, opened 2026-08-28
Create 03_fix_display_fr_file_classname.sql

Sebastienfavre:
Some class values are missing in French, whilst others are written entirely in capital letters : 
image


ponceta (approved):
Thank you!

